PR #3, opened 2019-02-26
Use the RBAC service for Add/Edit/Remove groups and list users

lgalis:
Use the RBAC service for Add/Edit/Remove groups and list users

screenshot from 2019-03-05 15-20-01
screenshot from 2019-03-05 15-20-09
screenshot from 2019-03-05 15-20-23
screenshot from 2019-03-05 15-20-42


package.json
```diff
@@ -56,7 +53,7 @@
     "@babel/preset-react": "^7.0.0",
     "babel-core": "7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
-    "babel-jest": "^23.6.0",
+    "babel-jest": "^24.1.0",
```
Hyperkid123: the 24 version of jest is a bit weird could you try to add some random snapshot test and see if its still working please?

src/Helpers/Group/GroupHelper.js
```diff
@@ -1,26 +1,42 @@
...
-const userApi = getApprovalApi();
+const groupApi = getGroupApi();
 
 export async function fetchGroups() {
-  let groups = await userApi.fetchGroups();
+  let groupsData = await groupApi.listGroups();
+  let groups = groupsData.data;
+  console.log('Groups: ', groups);
   let len = groups.length;
   for (let idx = 0; idx < len; idx++) {
```
Hyperkid123: I would use `map` or `forEach` it works with `async/await` just fine.

src/Helpers/Group/GroupHelper.js
```diff
@@ -1,26 +1,42 @@
...
+import { getGroupApi } from '../Shared/userLogin';
 
-const userApi = getApprovalApi();
+const groupApi = getGroupApi();
 
 export async function fetchGroups() {
-  let groups = await userApi.fetchGroups();
+  let groupsData = await groupApi.listGroups();
+  let groups = groupsData.data;
+  console.log('Groups: ', groups);
```
Hyperkid123: leftover console.log

src/Helpers/Group/GroupHelper.js
```diff
@@ -1,26 +1,42 @@
...
+    groups[idx].members = groupWithUsers.principals;
   }
+
   return groups;
 }
 
 export async function updateGroup(data) {
-  await userApi.updateGroup(data.id, data);
+  await groupApi.updateGroup(data.uuid, data);
+  let members_list = data.members.map(user => { return user.username; });
```
Hyperkid123: You can return the value without the block. Also i don't see the `members_list` variable ever change so you should use `const` instead of `let`. And linter will probably complain about not using camel calse
```javascript
const membersList = data.members.map(user => user.username);
```

src/Helpers/Group/GroupHelper.js
```diff
@@ -1,26 +1,42 @@
...
+  //update the user members here - adding users and removing users from the group should be a separate action in the UI
+  let addUsers = data.user_list.filter(item => !members_list.includes(item.username));
+  let removeUsers = members_list.filter(item => !(data.user_list.map(user => { return user.username;})).includes(item));
+  if (addUsers.length > 0) {
+    await groupApi.addPrincipalToGroup(data.uuid, JSON.stringify({ principals: addUsers }));
```
Hyperkid123: no need for `JSON.stringify` that should be done by default.

src/Helpers/Group/GroupHelper.js
```diff
@@ -1,26 +1,42 @@
...
+  }
+
+  if (removeUsers.length > 0) {
+    await groupApi.deletePrincipalFromGroup(data.uuid, removeUsers.join(','));
```
Hyperkid123: instead of `join` i would use template string its a bit safer n case
```javascript
await groupApi.deletePrincipalFromGroup(data.uuid, `${removeUsers}`);
```

src/Helpers/Group/GroupHelper.js
```diff
@@ -1,26 +1,42 @@
...
+  }
 }
 
 export async function addGroup(data) {
-  await userApi.addGroup(data);
+  let newGroup = await groupApi.createGroup(data);
+  groupApi.addPrincipalToGroup(newGroup.uuid, JSON.stringify({ principals: data.user_list }));
```
Hyperkid123: `JSON.stringify`

src/Helpers/Shared/userLogin.js
```diff
@@ -1,12 +1,30 @@
-let ApprovalApi = require('insights_service_approval_ap_is');
+import { RBAC_API_BASE, RBAC_USER, RBAC_PWD } from '../../Utilities/Constants';
 
-import { APPROVAL_API_BASE } from '../../Utilities/Constants';
+let RoleBasedAccessControl = require('role_based_access_control');
```
Hyperkid123: i would prefer `import` instead of `require`

src/Helpers/User/UserHelper.js
```diff
@@ -1,27 +1,26 @@
-import { getApprovalApi } from '../Shared/userLogin';
+import { getPrincipalApi } from '../Shared/userLogin';
 
-const userApi = getApprovalApi();
+const principalApi = getPrincipalApi();
+
+//export async function fetchUsersRbac() {
```
Hyperkid123: can we delete the commented code?

src/SmartComponents/AppTabs/apptabs.scss
```diff
@@ -2,6 +2,8 @@
 @import '~@red-hat-insights/insights-frontend-components/Utilities/_all';
 
 .ins-tabs {
-  padding-left:12px;
+  padding-left: 12px;
+  padding-right: 12px;
+  margin-bottom: 12px;
   background-color: white;
 }
```
Hyperkid123: Missing new line

src/SmartComponents/Group/add-group-modal.js
```diff
@@ -20,7 +20,7 @@ const AddGroupModal = ({
   updateGroup
 }) => {
   const onSubmit = data => {
-    data.user_ids = selectedUsers;
+    data.user_list = selectedUsers.map((user) => { return { username: user };});
```
Hyperkid123: You should never override function params instead copy it into new variable.
If you want to return object directly you can sue this notation instead
```javascript
const something = {
  ...data,
  user_list: selectedUsers.map(user =>  ({ username: user })
}
```

src/SmartComponents/Group/add-group-modal.js
```diff
@@ -40,7 +40,7 @@ const AddGroupModal = ({
   const onOptionSelect = (selectedValues = []) =>
   { selectedUsers = selectedValues.map(val => val.value); };
 
-  const dropdownItems = users.map(user => ({ value: user.id, label: `${user.first_name} ${user.last_name}`, id: user.id }));
+  const dropdownItems = users.map(user => ({ value: user.username, label: `${user.username}`, id: user.username }));
```
Hyperkid123: no need for template string anymore
```javascript
const dropdownItems = users.map(user => ({ value: user.username, label: user.username, id: user.username }));
```

src/SmartComponents/Group/add-group-modal.js
```diff
@@ -77,7 +77,8 @@ const AddGroupModal = ({
             isMulti={ true }
             placeholders={ 'Select Members' }
             options={ dropdownItems }
-            defaultValue={ initialValues ? initialValues.members.map(user => user.id) : [] }
+            defaultValue={ (initialValues && initialValues.members) ? initialValues.members.map(
```
Hyperkid123: You can get rid of this condition if you set `initialValues` in component `defaultProps`. Unless there is a case when initial values can be object without `members` property,

src/SmartComponents/User/User.js
```diff
@@ -36,7 +35,7 @@ class User extends Component {
   };
 
   buildUserActionKebab = (user) => {
-    console.log('isKebab open for ', user.name, this.state.isKebabOpen);
+    console.log('isKebab open for ', user.username, this.state.isKebabOpen);
```
Hyperkid123: leftover console.log

src/SmartComponents/User/User.js
```diff
@@ -85,7 +84,7 @@ class User extends Component {
         <DataListCheck aria-labelledby={ `check-user-${item.id}` } name={ `check-user-${item.id}` }/>
         <DataListCell>
           <StackItem>
-            <span id={ item.id }>{ `${item.first_name} ${item.last_name}` } </span>
+            <span id={ item.id }>{ `${item.username}` } </span>
```
Hyperkid123: No need for template strings

src/SmartComponents/User/Users.js
```diff
@@ -67,7 +55,7 @@ class Users extends Component {
         <Route exact path="/users/remove/:id" component={ RemoveUser } />
         <Section type='content'>
           { this.renderToolbar() }
-          <UserList { ...filteredItems } noItems={ 'No Approvers' } fetchGroupsByUserId = { this.props.fetchGroupsByUserId }/>
+          <UserList { ...filteredItems } noItems={ 'No Principals' } fetchGroupsByUserId = { this.props.fetchGroupsByUserId }/>
```
Hyperkid123: no spaces here `fetchGroupsByUserId={`

src/entry-dev.js
```diff
@@ -11,7 +11,7 @@ import App from './App';
  */
 ReactDOM.render(
   <Provider store={ store }>
-    <Router basename='/insights/platform/approval'>
+    <Router basename='/insights/platform/rbac'>
```
Hyperkid123: is basename required?

src/SmartComponents/User/User.js
lgalis: I left the template strings because it will eventually go back to first_name/last_name - but I will remove it and added it back when needed.

src/entry-dev.js
Hyperkid123: I guess we will need something like this in Catalog

src/Helpers/Group/GroupHelper.js
lgalis: I would keep the join for now, until the rbac service is more established with the principals management.

package.json
lgalis: I have tested with this version and do not seem to have any issues.

src/SmartComponents/Group/add-group-modal.js
lgalis: the groups can be created without the members property.

mkanoor:
@lgalis @Hyperkid123 @gmcculloug 
The RBAC team has changed to use limit and offset for pagination we might want to make similar changes 
https://github.com/RedHatInsights/insights-rbac/pull/29

src/entry-dev.js
lgalis: I will have the change for /${RELEASE}/platform in a separate PR. This is already big as it is.

lgalis:
@mkanoor - thank you!  I will add those changes on separate PR.

package.json
lgalis: will do

lgalis: Changing to 23.6 until the 24 becomes more stable.

Hyperkid123:
> The RBAC team has changed to use limit and offset for pagination we might want to make similar changes
> [RedHatInsights/insights-rbac#29](https://github.com/RedHatInsights/insights-rbac/pull/29)

@mkanoor We are using limit/offset already.

lgalis:
@Hyperkid123, @gmcculloug  - feedback addressed.
Please review/merge.

src/Helpers/Group/GroupHelper.js
```diff
@@ -6,13 +6,8 @@ const groupApi = getGroupApi();
...
   let groups = groupsData.data;
-  let len = groups.length;
-  for (let idx = 0; idx < len; idx++) {
-    let groupWithUsers = await groupApi.getGroup(groups[idx].uuid);
-    groups[idx].members = groupWithUsers.principals;
-  }
-
-  return groups;
+  return Promise.all(groups.map(async group => { let groupWithUsers = await groupApi.getGroup(group.uuid);
+    return { ...group, members: groupWithUsers.principals };}));
```
Hyperkid123: new line after `};`

PR #5, opened 2019-03-22
Fix lint errors and remove extra file

lgalis:
Fix lint errors and remove extra file


lgalis:
@gmcculloug - fix the lint errors, removed the unused file and added the snapshot files for testing

PR #8, opened 2019-04-08
Groups pagination

Hyperkid123:
### Bug fixes
- fixed spandx config
- fixed mismatching versions of `react` and `react-dom`
- fixed missing initial state from groupReducer

### Changes
- use axios based js client for groups
- reworked groups screen
  - use PF4 table instead of data-list
  - added pagination

There is also an issue with the rbac paginated response. Its missing pagination meta data. I've created issue here: https://github.com/RedHatInsights/insights-rbac/issues/55. using workraound until its fixed.

screenshot-ci foo redhat com-1337-2019 04 08-14-51-11


lgalis:
@Hyperkid123 - the missing pagination metadata was fixed in the rbac service

Hyperkid123:
I see i will update it

PR #9, opened 2019-04-08
 Add back the group edit/remove action kebab

lgalis:
- Added back the group edit/remove action kebab in the group table 
 - Fix handling of no users case - groups can be created/removed/updated
 - Change landing page to the Groups Tab
 - Fix updating of table rows on group edit/add/remove

Screenshot from 2019-04-09 01-14-21

### DIFF
--- a/src/SmartComponents/Group/Groups.js
+++ b/src/SmartComponents/Group/Groups.js
@@ -125,6 +125,25 @@ class Groups extends Component {
       );
     }
 
+    actionResolver = (groupData, { rowIndex }) => {
+      if (rowIndex === 1) {
+        return null;
+      }
+
+      return [
+        {
+          title: 'Edit',
+          onClick: (event, rowId, group) =>
+            this.props.history.push(`/groups/edit/${group.uuid}`)
+        },
+        {
+          title: 'Delete',
+          onClick: (event, rowId, group) =>
+            this.props.history.push(`/groups/remove/${group.uuid}`)
+        }
+      ];
+    };
+
     render() {
       return (
         <Fragment>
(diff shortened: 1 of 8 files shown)


gmcculloug:
@lgalis Please see failing tests.

@Hyperkid123 Please review

src/SmartComponents/Group/Groups.js
```diff
@@ -45,6 +45,12 @@ class Groups extends Component {
       scrollToTop();
     }
 
+    componentDidUpdate(prevProps) {
+      if (this.props.groups !== prevProps.groups) {
```
Hyperkid123: This is gonna trigger every time anything happens with props. in JS `[1,1] !== [1,1]`. Props are always immutable objects and arrays will be different instances so this its always `true`. Use `lodash.isEqual` to compare arrays or objects.

src/SmartComponents/Group/add-group-modal.js
```diff
@@ -40,7 +48,7 @@ const AddGroupModal = ({
   const onOptionSelect = (selectedValues = []) =>
   { selectedUsers = selectedValues.map(val => val.value); };
 
-  const dropdownItems = users.map(user => ({ value: user.username, label: user.username, id: user.username }));
+  const dropdownItems = users ? users.map(user => ({ value: user.username, label: user.username, id: user.username })) : [];
```
Hyperkid123: instead of this condition i would prefer adding default props to users 

```javascript
AddGroupModal.defaultProps = {
  users: []
}
```

src/redux/Actions/GroupActions.js
```diff
@@ -20,6 +20,13 @@ export const fetchUsersByGroupId = apiProps => ({
   })
 });
 
+export const fetchGroup = apiProps => ({
+  type: ActionTypes.FETCH_GROUP,
+  payload: new Promise(resolve => {
```
Hyperkid123: No need for the Promise wrapper. You can use the helper function straight away.

lgalis:
@Hyperkid123 - feedback addressed

PR #10, opened 2019-04-10
Refactoring

Hyperkid123:
### Changes
- removed unused JS dependencies
- fixed JS dependencies security vulnerabilities
- removed some unused assets
- fix the tabs spacing issue
- tabs are now activated based on current route
- remove sass includes from webpack because we were not using them


lgalis:
Adding a before and after screenshots for the tabs:

Before:

Screenshot from 2019-04-10 16-29-28

After:
Screenshot from 2019-04-10 16-31-00

PR #13, opened 2019-04-15
Ui styling

Hyperkid123:
### Changes
- fixed component styling
  - tabs spacing
  - top toolbar spacing
  - table toolbar spacing
- removed unused SCSS files and CSS classes

### Bugfixes
- removed groups fetch disable
  - the groups table was throwing errors if the async action was not dispatched
  - trying to call `then` on undefined promise

Will do more clean up in follow up PR

### Before
screenshot-ci cloud paas upshift redhat com-2019 04 15-14-33-56

### After
screenshot-ci foo redhat com-1337-2019 04 15-14-34-05

PR #14, opened 2019-04-15
Refactoring Users components

Hyperkid123:
Waiting for #13 

### Changes
- removed all unused User components and functionalities
  - adding/editing/deleting users
  - redux references to above functions
  - related scss files
- Code clean up for users components
- removed some unused config files
- converted .eslintrc to JSON so we can used official docs 
- downgraded PF libraries to latest stable release (was using unstable versions)

No visual changes here


Hyperkid123:
cc @lgalis @gmcculloug

PR #19, opened 2019-04-30
 Remove User components and the User tab

lgalis:
Remove calls for lists of users from the group components. 

No User tab:
Screenshot from 2019-04-30 16-37-03


lgalis:
@gmcculloug

gmcculloug:
@Hyperkid123 Please review.  At this point this UI will be reworked, and the principal list will be completely re-done.  So removing now as only limited users have access to the principal data.

PR #20, opened 2019-05-01
Update browser title to just the App name

lgalis:
Update browser title to just the App name

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,7 +2,7 @@
 <html lang="en-US">
     <head>
         <meta charset="UTF-8">
-        <title>RBAC</title>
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en-US">
     <head>
         <meta charset="UTF-8">
-        <title>Red Hat Cloud Services Software | RBAC Service</title>
+        <title>RBAC</title>
```
gmcculloug: Please update this to `User Access Management` to match the navigation.  Thanks.

lgalis: Done

PR #25, opened 2019-06-10
Allow adding a user (exact name)  to a group

lgalis:
Allow adding a user ( exact name)  to a group. 
Using the react select creatable for now, until a similar component is available in PF4. 
A user that is not an org-admin cannot verify if he typed user exist.

There is no way at the moment for an RBAC admin that is not an org admin to verify that the typed user is correct. When adding a mistyped user to the group, there is no error to indicate that the user is not added.
I will create a follow-up PR to verify that the added users match the user selection and notify the user.

Screenshot from 2019-06-12 17-36-41


gmcculloug:
@Hyperkid123 Please review

Hyperkid123:
I found some issues.

- when adding a group i see the select but cant add anything to it
- i am not able to remove users from group (when i click the chip or when i want to clear all)
- when editing groups with two or more users and i click the remove icon on the chip it duplicates the users
groups

lgalis:
@Hyperkid - please re-review

PR #26, opened 2019-06-15
Update file names to match convention

lgalis:
Update file names to match convention.

Should be merged after #25


lgalis:
@Hyperkid123  - please review

Hyperkid123:
@lgalis @gmcculloug this change seem ok but when i run the rbac UI locally i am getting redirected to catalog again. Are you aware of any possible breaking changes on our or platform side?

Hyperkid123:
@lgalis works. I believe that there is an outdated version PF. Can you update it (not in this PR)?

lgalis:
@Hyperkid123 - thank you for the review - I'll update the PF4 version

Issue #32, opened 2019-08-30
Add "Roles" tab


dlabrecq:
As a user, I want to see a tab to display RBAC roles. This should appear similar to the lo-fi mock below.

Mock: https://marvelapp.com/c34g5a7/screen/58499708

Mock screenshot:
<img width="1439" alt="Screen Shot 2019-08-30 at 10 28 16 AM" src="https://user-images.githubusercontent.com/17481322/64028582-e5fd2a80-cb10-11e9-9904-907db75e4aed.png">

Note that we will not implement edit and delete functionality. We will also use the Insights toolbar for now. PatternFly is in the process of refactoring the toolbar component.

PR #33, opened 2019-08-30
Added roles tab

dlabrecq:
Added a tab to display RBAC roles, per the lo-fi mock below.

Mock: https://marvelapp.com/c34g5a7/screen/58499708

Note that we will not implement edit and delete functionality. We will also use the Insights toolbar for now. PatternFly is in the process of refactoring the toolbar component.

Fixes https://github.com/RedHatInsights/insights-rbac-ui/issues/32

Screen Shot 2019-09-03 at 10 42 39 AM

### DIFF
--- a/src/helpers/role/role-helper.js
+++ b/src/helpers/role/role-helper.js
@@ -8,6 +8,18 @@ export async function fetchRoles({ limit, offset }) {
   return roles;
 }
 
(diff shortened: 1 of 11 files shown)


dlabrecq:
Refactored to use shared toolbar

PR #35, opened 2019-09-03
Add missing create policy action

lgalis:
Add missing create policy action.


lgalis:
@Hyperkid123 - please review. Will add tests to catch a rename like this.

PR #37, opened 2019-09-05
Add the missing roles path

lgalis:
Add the missing roles path - fixes navigation in chrome to the User Management Settings


gmcculloug:
@lgalis Thanks.

PR #53, opened 2019-10-09
Add policy edit ( information and roles modals)

lgalis:
Implements https://projects.engineering.redhat.com/browse/SSP-756

Screenshot from 2019-10-11 14-56-28

Screenshot from 2019-10-11 14-56-05


src/App.scss
```diff
@@ -29,6 +29,10 @@
   }
 }
 
+.select-style {
```
Hyperkid123: I do understand why this exists but i am not sure if this is the way forward (right now there is no other option...). This is problem with forms inside modals. I will bring this UP with Susan and try to figure out some future proof solution.

src/smart-components/group/policy/policy-actions/edit-policy-roles.js
```diff
@@ -0,0 +1,155 @@
...
+    if (policy) {
+      const policy_data = {
+        name: policy.name,
+        group: policy.group.uuid,
+        roles: selectedRoles.map(role => role.value)
+      };
+      return postMethod ? updatePolicy(policy.uuid, policy_data).then(() => postMethod()).then(() => push(closeUrl)) :
+        updatePolicy(policy.uuid, policy_data).then(() => push(closeUrl));
+    }
+    else {
```
Hyperkid123: no need fo the else branch. Main branch has return statement.

src/App.scss
lgalis: Sounds good, thank you!

lgalis: @Hyperkid123 - my next PR will replace the select with a compact table - that is in the UX mocks

src/smart-components/group/policy/policy-actions/edit-policy-roles.js
lgalis: @Hyperkid123  - Updated

PR #58, opened 2019-10-28
Added axios error interceptor

lgalis:
Added axios error interceptor


Hyperkid123:
@lgalis should all of these commits be in this PR?

The interceptor looks OK.

lgalis:
@Hyperkid123 - I pushed more commits by mistake - rebased  now, thank you.

PR #59, opened 2019-11-06
Hide add role if user is not entitled for cost management

karelhala:
### Changes

There's been a lot of redundant code and a lot of reundant calls to API. I've removed the code and streamlined role fetching. When user is not entitled to cost management no button will be shown. Also filter was not properly working, so I updated the fetch function to actually filter. Next step should be ordering and refactoring of groups table.

### UI

Screenshots taken on account without `cost_management` entitlements.

**Before**
Screenshot from 2019-11-06 16-42-20

**After - user not org admin or not on beta**
Screenshot from 2019-11-08 10-45-32

**After - user is org admin and on beta**
Screenshot from 2019-11-08 10-45-09

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "@babel/plugin-transform-runtime": "^7.2.0",
     "@babel/preset-env": "^7.2.3",
     "@babel/preset-react": "^7.0.0",
+    "axios-mock-adapter": "^1.17.0",
     "babel-core": "7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.7.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2861,6 +2861,15 @@
         }
       }
     },
+    "axios-mock-adapter": {
+      "version": "1.17.0",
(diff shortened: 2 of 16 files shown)


Hyperkid123:
@karelhala just fix the test please.

gmcculloug:
> When user is not entitled to cost management no button will be shown.

@karelhala Will this change affect Catalog usage?  Can you add a screenshot to clarify what is being changed?

karelhala:
@gmcculloug yup, that would effect Catalog as well, we have a JIRA RHCLOUD-2932 to disable custom roles for users not entitled to cost-management. Let me upload the pictures.

gmcculloug:
@karelhala Realizing this is not an issue for Catalog as we are not using custom roles.  👍

package.json
```diff
@@ -55,6 +55,7 @@
     "@babel/plugin-transform-runtime": "^7.2.0",
     "@babel/preset-env": "^7.2.3",
     "@babel/preset-react": "^7.0.0",
+    "axios-mock-adapter": "^1.17.0",
```
Hyperkid123: I think you can remove the `xhr-mock` from the list.

PR #64, opened 2019-11-20
Rbac add group

PanSpagetka:
Change Add group wizard according to mocks.
Remove Policies and add User table.
Screenshot from 2019-11-25 12-09-45
Screenshot from 2019-11-25 12-09-57
Screenshot from 2019-11-25 12-10-25
Screenshot from 2019-11-25 12-10-43

https://projects.engineering.redhat.com/browse/RHCLOUD-3328
https://marvelapp.com/1jfai5ig/screen/63628176


src/helpers/user/user-helper.js
```diff
@@ -4,6 +4,6 @@ const principalApi = getPrincipalApi();
 
 export async function fetchUsers() {
   let usersData = await principalApi.listPrincipals();
```
karelhala: You can return this await emediately
```JS
return await principalApi.listPrincipals();
```

src/smart-components/group/add-group/add-group-wizard.js
```diff
@@ -50,36 +49,27 @@ const AddGroupWizard = ({
...
+      component: new SetRoles({ formData, selectedRoles, setSelectedRoles, roles })
     },
     { name: 'Review',
       component: new SummaryContent({ values: formData, selectedUsers, selectedRoles }),
       nextButtonText: 'Confirm',
-      enableNext: isGroupInfoValid && isPolicyInfoValid
+      enableNext: isGroupInfoValid
     }
   ];
   const fetchData = () => {
```
karelhala: I would have change this function completely
```JS
const  fetchData = () => {
 fetchGroup(id);
 fetchRoles();
 fetchUsers()
}
```
You are already connected to redux store so why not use it. You'll also want to add users and roles to mapped props at the top of component definiton. Also you can remove `setRoles` and `setUsers`

src/smart-components/group/add-group/add-group-wizard.js
```diff
@@ -88,36 +78,12 @@ const AddGroupWizard = ({
...
-        })).then(() => push(closeUrl)) :
-        createPolicy(policy_data).payload.catch(() =>
-          addNotification({
-            variant: 'danger',
-            title: `Add group`,
-            dismissable: true,
-            description: `Error creating policy`
-          })).then(() => push(closeUrl));
+    await addGroup(user_data);
+    if (postMethod) {
```
karelhala: Instead of this you can add `postMethod: () => undefined` to `AddGroupWizard.defaultProps`

src/smart-components/group/add-group/users-list.js
```diff
@@ -0,0 +1,102 @@
...
+
+const columns = [
+  { title: 'User name', orderBy: 'name' },
+  { title: 'Email' },
+  { title: 'First name' },
+  { title: 'Last name' }
+];
+
+const createRows = (data, expanded, checkedRows = []) => {
+  console.log(data);
```
karelhala: Forgotten `console.log`

src/smart-components/group/add-group/users-list.js
```diff
@@ -0,0 +1,102 @@
...
+    fetchUsers({});
+  }, []);
+
+  const setCheckedItems = (newSelection) => {
+    setSelectedUsers((users) => {
+      return newSelection(users).map(({ uuid, name, label }) => ({ uuid, label: label || name }));
+    });
+  };
+
+  console.log('render', users);
```
karelhala: Another `console.log`

src/helpers/user/user-helper.js
Hyperkid123: This does not even has to be async/await function. the API already returns a promise

src/redux/actions/user-actions.js
```diff
@@ -7,3 +7,11 @@ export const fetchUser = () => ({
     resolve(UserHelper.fetchUser());
   })
 });
+
+export const fetchUsers = () => ({
+  type: ActionTypes.FETCH_USERS,
+  payload: new Promise(resolve => {
```
karelhala: Isn't `payload: UserHelper.fetchUsers()` enough? If not can you please add `await` and `async` to `fetchUsers` function.

src/helpers/user/user-helper.js
```diff
@@ -2,8 +2,6 @@ import { getPrincipalApi } from '../shared/user-login';
 
 const principalApi = getPrincipalApi();
 
-export async function fetchUsers() {
-  let usersData = await principalApi.listPrincipals();
-  let users = usersData.data;
-  return users;
+export function fetchUsers() {
+  return principalApi.listPrincipals();;
```
karelhala: double `;`

src/redux/actions/user-actions.js
PanSpagetka: Yep, `payload: UserHelper.fetchUsers()` is enough

PanSpagetka:
@karelhala now ii should be fine ;)

lgalis:
@PanSpagetka @karelhala - please see my comment above

PanSpagetka:
@lgalis Thanks for the comment. I have tried to add a group with user that is not an orgadmin and seems to be working. In the principal list user can see only himself, I think that is reasonable behavior. Could you please provide more detailed reproducer? However @karelhala is making PR where he is disabling button for non orgadmin users #77 so this shouldn't be issue.

I have to admit that I am little bit confused with what which should be able to see/do. We need to definitely clarify that. But adding users to a group based one exact match of username doesn't seem as a good solution to me. It can create security issues and UX is also not very nice. Until now, I didn't hear about such requirements. Could you tell me why do we need such a feature? Or from where this requirement comes from?

karelhala:
@lgalis thanks for pointing that out. We have released a new PR that hides add group button so users without org admin rights should not be able to add new group. That comes from our requirements as only power users should be able to do these things.

lgalis:
@PanSpagetka, @karelhala - that is a change from what there used to be, as users that are part of the RbacAdmin group would also be able to add users to a group, by providing the exact username. ( I know there are changes to the backend - so I will follow up to see what the new guidelines are).

PR #75, opened 2019-11-25
[WIP] Move Edit and Delete Groups

tahmidefaz:


tahmidefaz:
This is a work in progress, as of now. [RHCLOUD-3330](https://projects.engineering.redhat.com/browse/RHCLOUD-3330) doesn't really require dealing with the checkboxes, but I figured it would be better to have that in through this PR as it would make it easier to deal with the kebab in the table header. 
Currently scratching my head around why it throws `setSelectedRows is not a function` when I try to select a row. Would appreciate any suggestions.

PR #76, opened 2019-11-25
Update group kebabs

tahmidefaz:
Jira: [RHCLOUD-3330](https://projects.engineering.redhat.com/browse/RHCLOUD-3330)
Mockups: [Here](https://marvelapp.com/5dcjg6g/screen/63628276)

**Some things to note:**
- The acceptance criteria don't require dealing with the checkboxes. I figured it would be better to enable those for group edit/delete. I think the checkboxes will be updated in a future PR to align more closely with the UX mockups.
- The multiple delete group button doesn't really do anything atm, we are waiting on mockups here. Once that's done, we can add the modal there.

**Current Look**
image

Any suggestions would be greatly appreciated :slightly_smiling_face:

### DIFF
--- a/src/smart-components/group/groups.js
+++ b/src/smart-components/group/groups.js
@@ -4,7 +4,7 @@ import { bindActionCreators } from 'redux';
 import PropTypes from 'prop-types';
 import { Link, Route, Switch } from 'react-router-dom';
 import { expandable } from '@patternfly/react-table';
-import { Button, Stack, StackItem, DropdownItem } from '@patternfly/react-core';
+import { Button, Stack, StackItem } from '@patternfly/react-core';
 import AddGroupWizard from './add-group/add-group-wizard';
 import EditGroup from './edit-group-modal';
 import RemoveGroup from './remove-group-modal';
@@ -74,7 +74,7 @@ const Groups = ({ fetchGroups, isLoading, pagination, history: { push }, groups 
     {
       label: 'Edit group',
       props: {
-        isDisabled: selectedRows.length == 1 ? false : true
+        isDisabled: selectedRows.length === 1 ? false : true
       },
       onClick: () => push(`groups/edit/${selectedRows[0].uuid}`)
     },
--- a/src/smart-components/group/groups.js
+++ b/src/smart-components/group/groups.js
@@ -1,7 +1,7 @@
 import React, { Fragment, useState, useEffect } from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import PropTypes, { arrayOf } from 'prop-types';
+import PropTypes from 'prop-types';
 import { Link, Route, Switch } from 'react-router-dom';
 import { expandable } from '@patternfly/react-table';
 import { Button, Stack, StackItem } from '@patternfly/react-core';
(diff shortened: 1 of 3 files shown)


src/smart-components/group/groups.js
```diff
@@ -34,6 +35,12 @@ const Groups = ({ fetchGroups, isLoading, pagination, history: { push }, groups
     fetchGroups(config);
   };
 
+  const setCheckedItems = (newSelection) => {
+    setSelectedRows((rows) => {
+      return newSelection(rows).map(({ uuid, name }) => ({ uuid, label: name }));
```
Hyperkid123: no need for new code block:
```jsx
setSelectedRows((rows) => newSelection(rows).map...)
```

src/smart-components/group/groups.js
```diff
@@ -62,9 +68,22 @@ const Groups = ({ fetchGroups, isLoading, pagination, history: { push }, groups
...
       >
-        Add group
+        Create a group
       </Button>
-    </Link>
+    </Link>,
+    {
+      label: 'Edit group',
+      props: {
+        isDisabled: selectedRows.length === 1 ? false : true
```
Hyperkid123: `selectedRows.length === 1` is enough. It will always return boolean value

src/smart-components/group/groups.js
```diff
@@ -62,9 +68,22 @@ const Groups = ({ fetchGroups, isLoading, pagination, history: { push }, groups
...
+      },
+      onClick: () => push(`groups/edit/${selectedRows[0].uuid}`)
```
Hyperkid123: I think you need the leading `/` in 
```jsx
push(`/groups/edit/${selectedRows[0].uuid}`)
```
You want to replace the whole path not append something to it.

src/smart-components/group/groups.js
```diff
@@ -81,6 +100,9 @@ const Groups = ({ fetchGroups, isLoading, pagination, history: { push }, groups
             data={ groups }
             createRows={ createRows }
             columns={ columns }
+            isSelectable={ true }
```
Hyperkid123: `isSelectable={true}` is equivalent to `isSelectable`. This should be caught by linter.

src/smart-components/group/groups.js
```diff
@@ -125,6 +147,8 @@ Groups.propTypes = {
   isLoading: PropTypes.bool,
   searchFilter: PropTypes.string,
   fetchGroups: PropTypes.func.isRequired,
+  selectedRows: PropTypes.array,
```
Hyperkid123: I always prefer using `arrayOf(PropTypes.shape({...}))`. It makes the type checking more secure.

tahmidefaz:
@Hyperkid123 Thanks for the feedback. Looks like I was able to incorporate all your suggestions

src/smart-components/group/groups.js
```diff
@@ -125,6 +145,14 @@ Groups.propTypes = {
   isLoading: PropTypes.bool,
   searchFilter: PropTypes.string,
   fetchGroups: PropTypes.func.isRequired,
+  selectedRows: arrayOf(
```
karelhala: `selectedRows` and `setSelectedRows` comes from state, so no need to add them to propTypes at all.

tahmidefaz: done! :slightly_smiling_face:

PR #77, opened 2019-11-26
When on groups page hide add group for non org admins

karelhala:
## UI changes
* Users without org admin rights should not be able to add new group

### Before
Screenshot from 2019-11-26 10-54-34

### After
Screenshot from 2019-11-26 10-54-17

#### JIRA
RHCLOUD-3532

### DIFF
--- a/src/helpers/group/group-helper.js
+++ b/src/helpers/group/group-helper.js
@@ -3,7 +3,10 @@ import { getGroupApi } from '../shared/user-login';
 const groupApi = getGroupApi();
 
 export async function fetchGroups({ limit, offset, name, orderBy }) {
-  return await groupApi.listGroups(limit, offset, name, orderBy);
+  return {
+    ...await groupApi.listGroups(limit, offset, name, orderBy),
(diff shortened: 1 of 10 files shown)


tahmidefaz:
@karelhala you might want to make these changes over #76 :slightly_smiling_face:

src/helpers/group/group-helper.js
```diff
@@ -4,7 +4,10 @@ import { getGroupApi } from '../shared/user-login';
 const groupApi = getGroupApi();
 
 export async function fetchGroups({ limit, offset, name, orderBy }) {
-  return await groupApi.listGroups(limit, offset, name, orderBy);
+  return {
+    ...await groupApi.listGroups(limit, offset, name, orderBy),
```
Hyperkid123: is this going to be sequential?

karelhala: Yeah, good ponit! I've changed it to use `Promise.all` so these promises don't wait for each other.

PR #79, opened 2019-11-27
Add roles to Group detail

PanSpagetka:
Add roles to Group detail
RBAC api js client is not updated yet. So the Role related API calls will result in error. Instead of displaying list of roles for group are currently displayed all roles.

https://projects.engineering.redhat.com/browse/RHCLOUD-3332


PanSpagetka:
@karelhala review please

src/helpers/group/group-helper.js
```diff
@@ -46,3 +45,17 @@ export async function deletePrincipalsFromGroup(groupId, users) {
 export async function addPrincipalsToGroup(groupId, users) {
   return await groupApi.addPrincipalToGroup(groupId, { principals: users });
 }
+
+export async function fetchRolesForGroup(groupId) {
+
```
karelhala: Extra whitespace.

PR #123, opened 2020-02-04
Edit group

PanSpagetka:
https://projects.engineering.redhat.com/browse/RHCLOUD-4104
Screenshot from 2020-02-04 14-23-02
Screenshot from 2020-02-04 14-22-50
Screenshot from 2020-02-05 11-29-58

### DIFF
--- a/src/smart-components/group/group.js
+++ b/src/smart-components/group/group.js
@@ -1,4 +1,4 @@
-import React, { Fragment, useEffect } from 'react';
+import React, { Fragment, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Route, Switch, Redirect } from 'react-router-dom';
 import { connect } from 'react-redux';
@@ -9,6 +9,9 @@ import GroupPrincipals from './principal/principals';
 import GroupRoles from './role/group-roles';
 import { fetchGroup } from '../../redux/actions/group-actions';
 import { ListLoader } from '../../presentational-components/shared/loader-placeholders';
+import { Button } from '@patternfly/react-core';
+import EditGroup from './edit-group-modal';
+import './group.scss';
 
 const Group = ({
   match: { params: { uuid }},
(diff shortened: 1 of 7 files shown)


src/smart-components/group/group.scss
```diff
@@ -0,0 +1,3 @@
+.edit-group-button {
+  float: right;
```
karelhala: You shouldn't use floats, if you want to position the edit button you should use [level](https://patternfly-react.surge.sh/documentation/react/layouts/level) layout.

src/smart-components/group/group.js
```diff
@@ -37,6 +41,14 @@ const Group = ({
   return (
     <Fragment>
       <TopToolbar breadcrumbs={ breadcrumbsList() }>
+        <Button className='edit-group-button' onClick={ () => setShowEdit(true) } variant='secondary'>Edit Group</Button>
+        <EditGroup
+          isOpen={ showEdit }
+          group={ group }
+          closeUrl={ `group/detail/${uuid}` }
+          onClose={ () => setShowEdit(false) }
+          postMethod={ () => new Promise(() => setShowEdit(false)) }
```
karelhala: No need to wrap in new Promise. Once promise is initiated it will always be thenable, this can be only `setShowEdit(false)`.

However I think that we should fetch current group once the edit is finished so we refresh the UI and let user know about these changes.

src/smart-components/group/group.scss
karelhala: Can you remove this file? Looks like it is no longer needed.

mmenestr:
@PanSpagetka Hey Robin, thanks for including the screenshots.
Just a few comments:
1. The word "information" is misspelt in the header --> should be "information" NOT "informaiton"
2. The cancel button should use a **link** button instead of a **secondary** button!

Thank you!!

PR #124, opened 2020-02-04
show kebab in all rows

tahmidefaz:
This fixes the kebabs only showing up in alternating rows in group lists.
image


src/smart-components/group/groups.js
```diff
@@ -144,7 +144,9 @@ const mapDispatchToProps = dispatch => bindActionCreators({
 Groups.propTypes = {
   userIdentity: PropTypes.shape({
     user: PropTypes.shape({
+      /*eslint-disable camelcase*/
```
karelhala: If you want to disable rule for next line, you can go with `eslint-disable-next-line`

tahmidefaz: I will update it in a future pr! thanks!

PR #126, opened 2020-02-05
Add user tab

PanSpagetka:
Add user tab.

There is no `Last modified` column, because we don't get this information from API. I think we don't need this because we can't edit users.

Filtering is not working.
Bottom pagination line is not present, because we don't have total count yet. This should be updated when API provides us total count.
Screenshot from 2020-02-07 12-36-32

### DIFF
--- a/src/presentational-components/shared/top-toolbar.js
+++ b/src/presentational-components/shared/top-toolbar.js
@@ -38,7 +38,7 @@ export const TopToolbarTitle = ({ title, description, children }) => (
           { <Text component={ TextVariants.h1 }>{ title || <ToolbarTitlePlaceholder /> }</Text> }
         </TextContent>
         { description &&
-          <TextContent className="pf-u-pt-sm pf-u-mb-md">
+          <TextContent className="pf-u-pt-sm pf-u-mb-md top-toolbar-description">
             <Text component={ TextVariants.p }>{ description }</Text>
           </TextContent> }
       </LevelItem>
--- a/src/smart-components/user/users.js
+++ b/src/smart-components/user/users.js
@@ -1,0 +1,43 @@
+import React, {  useState } from 'react';
+import {  Stack, StackItem } from '@patternfly/react-core';
+import { TopToolbar, TopToolbarTitle } from '../../presentational-components/shared/top-toolbar';
+import { Section } from '@redhat-cloud-services/frontend-components';
+import UsersList from '../group/add-group/users-list';
+
+const Users = () => {
+  const [ selectedRows, setSelectedRows ] = useState([]);
+
+  const renderUsersList = () =>
+    <Stack>
+      <StackItem>
+        <TopToolbar paddingBottm={ false }>
+          <TopToolbarTitle title="Users" />
+        </TopToolbar>
+      </StackItem>
+      <StackItem>
+        <Section type="content" id={ 'users' }>
+          <UsersList
+            selectedUsers={ selectedRows }
+            setSelectedUsers={ setSelectedRows }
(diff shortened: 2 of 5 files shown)


mmenestr:
@PanSpagetka just a few comments:
1. For the header in the first column **"Username" should be one word,** not "User name" with a space in between.
2. Is it possible to add a description? as seen in the mocks?
3. The spacing for the header is a little off
- without a description, the header should have 24px of spacing above and below it
- with a description, the header should have 24px above it, 8px between the header and description, and 24px spacing right underneath it (before the start of the grey part of the page)
can see in more detail here [](https://marvelapp.com/4bf315h/screen/66016993)
(The blue/green rectangle means 24px spacing, and the bright green rectangle means 8px spacing)!

PanSpagetka:
@mmenestr I have fixed issues you have mentioned and updated the screenshot. Just one more thing, do we have the URL for `user management list` link?

karelhala:
The `user management list` should land users in https://www.qa.redhat.com/wapps/ugc/protected/usermgt/userList.html where `qa` part is only for ci and qa envs.

src/presentational-components/shared/top-toolbar.js
```diff
@@ -38,7 +38,7 @@ export const TopToolbarTitle = ({ title, description, children }) => (
           { <Text component={ TextVariants.h1 }>{ title || <ToolbarTitlePlaceholder /> }</Text> }
         </TextContent>
         { description &&
-          <TextContent className="pf-u-pt-sm pf-u-mb-md">
+          <TextContent className="pf-u-pt-sm pf-u-mb-md top-toolbar-description">
```
karelhala: Can you change the class to `ins-c-rbac__users-toolbar-description` are the `-` and `__` correct @ryelo?

src/smart-components/user/users.js
```diff
@@ -0,0 +1,46 @@
...
+const Users = () => {
+  const [ selectedRows, setSelectedRows ] = useState([]);
+
+  const renderUsersList = () =>
+    <Stack>
+      <StackItem>
+        <TopToolbar paddingBottm={ false }>
+          <TopToolbarTitle
+            title="Users"
+            description='These are the users in your organization. To add new users to your organization go to user management list.'
```
karelhala: Double quotes in here.

src/smart-components/user/users.js
```diff
@@ -0,0 +1,46 @@
+import React, {  useState } from 'react';
+import {  Stack, StackItem } from '@patternfly/react-core';
+import { TopToolbar, TopToolbarTitle } from '../../presentational-components/shared/top-toolbar';
+import { Section } from '@redhat-cloud-services/frontend-components';
+import UsersList from '../group/add-group/users-list';
+
+const Users = () => {
+  const [ selectedRows, setSelectedRows ] = useState([]);
```
karelhala: If we are not going to use the state in here, no need to pretend it like this. Rather use `() => undefined` in `setSelectedUsers` prop.

src/smart-components/user/users.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const renderUsersList = () =>
```
karelhala: No need to assign this to function call, but rather return this directly.

mmenestr:
@PanSpagetka yes, like Karel said, this is the link to the user management page https://www.redhat.com/wapps/ugc/protected/usermgt/userList.html

Also, one small thing to fix. The header and the description are not left aligned with each other. The description should be a few pixels to the left to be aligned!

PR #129, opened 2020-02-05
Delete modals enhancement

tahmidefaz:
Sorry, I had to re-write most of the existing delete group component, since we are enhancing it. 
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-4105
Mocks: https://marvelapp.com/a2263dd/screen/65614797

It does have one bug, though... when I delete multiple groups, it doesn't refresh the group list properly. For example, if I have two groups namely `000` and `001`, and I want to delete them using the new multiple group delete feature, it will delete them in reality. However, `001` will still show up when the modal closes, even though it doesn't exist anymore. It goes away when I refresh the page. Not sure what's causing this, let me know if you have any suggestions.


mmenestr:
Looks great! Just one thing, button for the modal where you are deleting multiple groups should say "Delete groups" with an s, since it is multiple groups :)

src/smart-components/group/remove-group-modal.js
```diff
@@ -1,89 +1,112 @@
...
+  const [ checked, setChecked ] = useState(false);
+
+  const multipleGroups = groups.length > 1;
+
+  const onSubmit = () => {
+    if (multipleGroups) {
+      let uuids = groups.map((group) => group.uuid);
+      removeGroups(uuids).then(() => postMethod()).then(push(closeUrl));
+    } else {
+      removeGroup([ groups[0].uuid ]).then(()=> postMethod()).then(push(closeUrl));
```
karelhala: I would have use `removeGroups` in here so we have on function to remove any amount of groups. Also, you probably forgot to pass UUIDs to postMethod. Also no need to use let if you are not reassigning it.
```JS
if (multipleGroups) {
  const uuids = groups.map((group) => group.uuid);
  removeGroups(uuids).then(() => postMethod(uuids)).then(push(closeUrl));
} else {
  removeGroups([ groups[0].uuid ]).then(([ groups[0].uuid ])=> postMethod()).then(push(closeUrl));
}
```

tahmidefaz: Thanks! I removed that whole conditional statement!

src/helpers/group/group-helper.js
```diff
@@ -36,8 +36,8 @@ export async function addGroup(data) {
   return ret;
 }
 
-export async function removeGroup(groupId) {
-  return await groupApi.deleteGroup(groupId);
+export async function removeGroups(uuids) {
+  return await uuids.map((uuid) => groupApi.deleteGroup(uuid));
```
karelhala: You should wait for all deletes, that will also fix the problem with refreshing
```JS
return await Promise.all( uuids.map((uuid) => groupApi.deleteGroup(uuid)));
```

tahmidefaz: Thanks, that fixed it! Trying to look into the tests now

tahmidefaz:
Thanks! ^ that commit fixes the lint

tahmidefaz:
Not sure what's going on with the test yet. It's not rendering the `<RemoveGroupModal/>` at all. Like when doing a `wrapper.debug()`, I am seeing something like:
```
<Router history={{...}}>
         <Route path="/groups/removegroups" render={[Function: render]} />
 </Router>
```
that `render={[Function: render]}` doesn't seem to execute at all.
Tried different things (even multiple `shallow()` and `dive()` methods), without any luck. I will give it another try tomorrow.

karelhala:
@tahmidefaz I've pulled your PR and looks like you have older version of node_modules installed. Can you run `npm install` and the running tests with updating snapshots? That at least fixes the issue with broken snapshots. I'll investigate further on why those 2 tests are failing entirely.

karelhala:
@tahmidefaz about the failing test that it can't find `Button` you are using incorrect initial route. Add `'/groups/removegroups'` to `initialEntries` and use 3rd `initialIndex` this way the UI will simulate landing on group removal URL where route with `RemoveGroupModal` is located. You will be able to find elements and test them correctly.

src/test/smart-components/group/remove-group-modal.test.js
```diff
@@ -11,24 +11,26 @@ import { RBAC_API_BASE } from '../../../utilities/constants';
...
   let initialState;
 
-  const GroupWrapper = ({ store, children }) => (
+  const GroupWrapper = ({ store }) => (
     <Provider store={ store }>
       <MemoryRouter initialEntries={ [ '/groups/', '/groups/123/', '/groups/' ] } initialIndex={ 1 }>
-        { children }
+        <Route path="/groups/removegroups" render={ (args) => <RemoveGroupModal { ...args } { ...initialProps }
+          isModalOpen
+          groupsUuid={ [{ uuid: '1234' }] } /> } />
```
karelhala: This should be the same as what you are expecting in mock API. So `123` based on the rest of the code around.

tahmidefaz: Good catch! Totally missed it!

tahmidefaz:
@karelhala Thanks a lot! I have updated the tests with your suggestions. Got both of them to pass locally!

PR #136, opened 2020-02-10
Fix info text when deleting single role form group

PanSpagetka:
Modal was showing link instead of plain text. Also it was using plural form in title for one selected user.

#### Before
Screenshot from 2020-02-10 15-24-32
#### After
Screenshot from 2020-02-10 15-39-09

### DIFF
--- a/src/smart-components/group/role/group-roles.js
+++ b/src/smart-components/group/role/group-roles.js
@@ -131,7 +131,7 @@ const GroupRoles = ({
             const multipleRolesSelected = selectedRoles.length > 1;
             setConfirmDelete(() => () => removeRoles(uuid, selectedRoles.map(role => role.uuid), () => fetchRolesForGroup(pagination)(uuid)));
             setDeleteInfo({
-              title: 'Remove roles?',
+              title: multipleRolesSelected ? 'Remove roles?' : 'Remove role?',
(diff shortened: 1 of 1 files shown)


src/smart-components/group/role/group-roles.js
```diff
@@ -131,9 +131,12 @@ const GroupRoles = ({
             const multipleRolesSelected = selectedRoles.length > 1;
             setConfirmDelete(() => () => removeRoles(uuid, selectedRoles.map(role => role.uuid), () => fetchRolesForGroup(pagination)(uuid)));
             setDeleteInfo({
-              title: 'Remove roles?',
+              title: multipleRolesSelected ? 'Remove roles?' : 'Remove role?',
               confirmButtonLabel: selectedRoles.length > 1 ? 'Remove roles' : 'Remove role',
-              text: removeModalText(name, multipleRolesSelected ? selectedRoles.length : selectedRoles[0].label, multipleRolesSelected)
+              text: removeModalText(
+                name,
+                multipleRolesSelected ? selectedRoles.length : roles.find(role => role.uuid === selectedRoles[0].uuid).name,
```
karelhala: You should be able to use directly `selectedRoles` name in here
```JS
multipleRolesSelected ? selectedRoles.length : selectedRoles[0].name,
```

PR #141, opened 2020-02-11
Add permissions count to roles list

karelhala:
### Before
Screenshot from 2020-02-11 13-52-21
Screenshot from 2020-02-11 13-52-26

### After
Screenshot from 2020-02-11 16-22-07
Screenshot from 2020-02-11 16-22-16


karelhala:
ping @mmenestr @KendraMar

mmenestr:
Thanks for tagging us Karel!! Here are my comments:

_**Comments for main page:**_

- "Role" column header should be called “Name” (I think we had talked about how the filter would be retrieving the value "name" not "role"? but let me know if that's wrong)
- All content in the roles column should be left aligned with the header, and same thing for the permissions column.
- **3rd column header should be “Permissions” not “Policies” !!**

_**Comments for permissions page:**_
- Question about the filter, what would they have to input for it to work? Would they be able to filter by any of the columns in the table?

karelhala:
> Thanks for tagging us Karel!! Here are my comments:
> 
> _**Comments for main page:**_
> 
> * "Role" column header should be called “Name” (I think we had talked about how the filter would be retrieving the value "name" not "role"? but let me know if that's wrong)

Sure thing! I'm on it.

> * All content in the roles column should be left aligned with the header, and same thing for the permissions column.

I checked and all columns are left aligned, except the `Permissions` column. That is a button, would you prefer it to be just a href without button?

> * **3rd column header should be “Permissions” not “Policies” !!**

It is `Permissions` I uncluded screenshots how it looked before (with `Policies`) and after this change as well.

> 
> _**Comments for permissions page:**_
> 
> * Question about the filter, what would they have to input for it to work? Would they be able to filter by any of the columns in the table?

As of right now users can input anything and it will filter out based on app, resource type and operations. Also if any of these values has `*` it will be included in the output as well. This is because we are doing the heavy lifting on frontend, because we don't really have endpoint for these permissions yet.

mmenestr:
> I checked and all columns are left aligned, except the Permissions column. That is a button, would you prefer it to be just a href without button?

Then yes, could you do it as a href without button! I know it's aligned in the back end but want it to be aligned visually too haha


> It is Permissions I included screenshots how it looked before (with Policies) and after this change as well.

and sorry I looked at the wrong one for the permissions column :D

> As of right now users can input anything and it will filter out based on app, resource type and operations. Also if any of these values has * it will be included in the output as well. This is because we are doing the heavy lifting on frontend, because we don't really have endpoint for these permissions yet.

- and great about the link!! thank you.

mmenestr:
@karelhala OH one more thing, "Last Modified" column header should be "Last modified", with lower case m!

karelhala:
@mmenestr these are the updated visuals
Screenshot from 2020-02-11 16-22-07
Screenshot from 2020-02-11 16-22-16

PR #147, opened 2020-02-13
Change filter by groups to filter by name

karelhala:
This PR changes the filter placeholder in groups, align first cell to left and update bottom padding to be aligned with mocks.

Screenshot from 2020-02-13 16-51-08

### JIRA RHCLOUD-4314

### DIFF
--- a/src/presentational-components/shared/top-toolbar.js
+++ b/src/presentational-components/shared/top-toolbar.js
@@ -8,7 +8,7 @@ import RbacBreadcrumbs from './breadcrubms';
 import './top-toolbar.scss';
 
 export const TopToolbar = ({ children,  breadcrumbs, paddingBottom }) => (
-  <div className={ `pf-u-pt-xl pf-u-pr-xl pf-u-pl-lg ${paddingBottom ? 'pf-u-pb-sm' : ''} top-toolbar` }>
+  <div className={ `pf-u-pt-xl pf-u-pr-xl pf-u-pl-lg ${paddingBottom ? 'pf-u-pb-md' : ''} top-toolbar` }>
     { breadcrumbs && (<Level className="pf-u-mb-md">
       <RbacBreadcrumbs { ...breadcrumbs } />
     </Level>)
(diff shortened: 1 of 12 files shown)


karelhala:
ping @mmenestr

src/presentational-components/shared/top-toolbar.js
```diff
@@ -8,7 +8,7 @@ import RbacBreadcrumbs from './breadcrubms';
 import './top-toolbar.scss';
 
 export const TopToolbar = ({ children,  breadcrumbs, paddingBottom }) => (
-  <div className={ `pf-u-pt-xl pf-u-pr-xl pf-u-pl-lg ${paddingBottom ? 'pf-u-pb-sm' : ''} top-toolbar` }>
+  <div className={ `pf-u-pt-xl pf-u-pr-xl pf-u-pl-lg ${paddingBottom ? 'pf-u-pb-md' : ''} top-toolbar` }>
```
Hyperkid123: `classNames`

mmenestr:
@karelhala Looks amazing spacing and alignment wise, thank you!! 

Just have 2 things I noticed:
- The description column should be a column titled "Roles" that shows the number of roles in each group, like in the mocks! Is that data retrievable?

- The default user access group has a checkbox and kebab, but that row should not have either! 
Is it possible to have **just that row** not have a bulk select and kebab? Also is it possible to pin that row to always be at the top of the first page? Let me know!

karelhala:
@mmenestr yeah, roles count is doable. I'll update it. About the default group, looks like checkbox can't be turned off (I'll file a PR in patternfly table to fix it), but I can hide the kebab there.

mmenestr:
@karelhala Ok, yes please do! and yes could you hide the kebab, and then disable the checkbox then just for that row?

karelhala:
I've made the checkbox disapear, it's a bit hackish way. Also about the default group being first, I am afraid we don't have a way how to show it there.

Screenshot from 2020-02-13 16-51-08

PR #148, opened 2020-02-13
Convert groups class to hook

fhlavac:
- replaced `connect()` functions with `useSelector()` and `useDispatch()`
- replaced history prop in groups with `useHistory()`
- removed redundant code
- updated tests and enhanced with pagination and filtering tests


src/smart-components/group/groups.js
```diff
@@ -1,49 +1,65 @@
...
   const routes = () => <Fragment>
-    <Route exact path="/groups/add-group" render={ props => <AddGroupWizard { ...props } postMethod={ fetchData } /> } />
-    <Route exact path="/groups/edit/:id" render={ props => <EditGroup { ...props } postMethod={ fetchData } isOpen/> } />
-    <Route exact path="/groups/removegroups" render={ props => <RemoveGroup { ...props } postMethod={ (ids) => {
-      fetchData();
-      setSelectedRows(selectedRows.filter(row => (!ids.includes(row.uuid))));
-      setFilterValue('');
-    } } isModalOpen groupsUuid={ removeGroupsList } /> } />
+    <Route exact path="/groups/add-group"
+      render={ props => <AddGroupWizard { ...props } postMethod={ config => dispatch(fetchGroups(config)) } /> }
```
Hyperkid123: While you are doing this, you should use the children syntax
```jsx
<Route>
  <AddGroupWizard postMethod={ config => dispatch(fetchGroups(config)) } />
</Route>
```
The `{...props}` should be replaced inside the component with hooks like `useHistory`, `useRouteMatch`, etc.

src/smart-components/group/groups.js
```diff
@@ -1,49 +1,65 @@
...
 const columns = [{ title: 'Name', cellFormatters: [ expandable ]}, 'Description', 'Members', 'Last modified' ];
 
-const Groups = ({ fetchGroups, isLoading, pagination, history: { push }, groups, userIdentity }) => {
+const Groups = () => {
   const [ filterValue, setFilterValue ] = useState('');
   const [ selectedRows, setSelectedRows ] = useState([]);
   const [ removeGroupsList, setRemoveGroupsList ] = useState([]);
 
+  const dispatch = useDispatch();
+  const { groups, pagination, userIdentity, isLoading } = useSelector(({ groupReducer: { groups, isLoading }}) => ({
```
karelhala: Can you add default value here as well? So it doesn't break for no reason.
```JS
const { groups, pagination, userIdentity, isLoading } = useSelector((
  { groupReducer: { groups, isLoading }} = { groupReducer: {} }
) => ({
```

Hyperkid123: I would solve this on the groupReducer level. There should always be an initial state.

fhlavac: Currently I have
```
// Initial State
export const groupsInitialState = {
  groups: {
    data: [],
    meta: {
      count: 0,
      limit: 10,
      offset: 0
    }
  },
  selectedGroup: { addRoles: {}},
  isLoading: false,
  isRecordLoading: false
};
```
in group-reducer.js, so it should not fail, should it? There are default values for everything.

Hyperkid123: Then there should be no problem

src/smart-components/group/add-group/add-group-wizard.js
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { withRouter } from 'react-router-dom';
+import { withRouter, useHistory } from 'react-router-dom';
```
Hyperkid123: Is anything from with router still used?

src/smart-components/group/remove-group-modal.js
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import { withRouter } from 'react-router-dom';
+import { withRouter, useHistory } from 'react-router-dom';
```
karelhala: Same as with `add-group-wizard.js` this component no longer uses anything from decorator `withRouter`.

PR #156, opened 2020-02-18
Create Group summary visual fix

PanSpagetka:
Display roles and users in columns under each other. Also I change compact navigation, so `Group  description` fits one line.
Screenshot from 2020-02-20 11-17-19
Screenshot from 2020-02-20 11-16-46

### DIFF
--- a/src/smart-components/group/add-group/summary-content.js
+++ b/src/smart-components/group/add-group/summary-content.js
@@ -21,7 +21,7 @@ const SummaryContent = (formData) => {
   return (
     <Fragment>
       <Stack gutter="md">
-        <StackItem>
+        <StackItem className="pf-u-mb-md">
           <Title size="xl"> Review </Title>
         </StackItem>
         <StackItem>
--- a/src/smart-components/group/add-group/summary-content.js
+++ b/src/smart-components/group/add-group/summary-content.js
@@ -55,9 +55,9 @@ const SummaryContent = (formData) => {
                 </GridItem>
                 <GridItem span= { 10 }>
                   <Text
-                    className="groups-table-detail content"
+                    className="groups-table-detail content pf-u-mb-md"
                     component={ TextVariants.h5 }>
-                    { `${selectedUsers.map((user, index) => `${index !== 0 ? ' ' : ''}${user.label}`)}` }
+                    { selectedUsers.map((role, index) => <Text className="pf-u-mb-0" key={ index }>{ role.label }</Text>) }
                   </Text>
                 </GridItem>
               </Grid>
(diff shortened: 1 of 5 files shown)


ryelo:
Looking at: https://marvelapp.com/1jfai5ig/screen/63628186/handoff

There are a few differences:
* sub header isn't bold and spacing is wrong
* sub header has the wrong text
* It looks like there needs to be a gutter on the items

ryelo:
also tagging @mmenestr

mmenestr:
@PanSpagetka @ryelo In this mockup, the formatting isn't necessarily "right" in terms of bolding. So I would go with the same bolding you've been using in the other steps of the wizards and go with that, for consistency. From what i've seen it seems like it is bolded everywhere else - so what I see in the screenshot is right.

The one thing I do notice though is that the spacing is inconsistent from "item" to "item". I would make the spacing between "group name" and "group description", and "group description" and "members", the same as the spacing between "members" and "roles". Looks like its about 16px?

I also would make the spacing between the header "Review" and the description way smaller... maybe also about 16px - or whatever it is in other wizard pages.

PanSpagetka:
@mmenestr 
- spacing between items should be 16px now.
- spacing between header and description is also 16px
- updated screenshots

karelhala:
Looks like snapshots needs to be updated as well.

mmenestr:
Discussed with Robin, this is the takeaway (this is based on looking at other wizards like sources for consistency):

Just go with wizard component defaults.
- Header should be bold
- Description should **not** be bold
- Spacing should be:
**Header**
24px
**Description**
24px
**Group name**
16px
**Group description**
16px
**Members**
16px
**Roles**

Also, change description to match mocks --> Confirm the details for this group, or click Back to revise.

src/smart-components/group/add-group/summary-content.js
```diff
@@ -35,41 +35,41 @@ const SummaryContent = (formData) => {
             <StackItem>
```
ryelo: instead of adding the `pf-u-mb-md` to each of the `Text` components, can we just wrap all of this in another `<Stack>` and then add a gutter?

src/smart-components/group/add-group/summary-content.js
```diff
@@ -21,7 +21,7 @@ const SummaryContent = (formData) => {
   return (
     <Fragment>
       <Stack gutter="md">
-        <StackItem>
+        <StackItem className="pf-u-mb-md">
           <Title size="xl"> Review </Title>
         </StackItem>
```
ryelo: We should update this to be similar to the previous pages
```
          <StackItem>
            <TextContent>
              <Title headingLevel="h4" size="xl"> {Title } </Title>
              <Text className="pf-u-mt-0" component={ TextVariants.h6 }> { description } </Text>
            </TextContent>
          </StackItem>
```

PanSpagetka:
@ryelo @mmenestr  
- Review page should be more similar to Assign roles/Add members. (Spacing, boldness, code).
- Description now use the right words
- Utility class was replaced for one css class used on upper element
- Updated snapshots

mmenestr:
Perfect, looks amazing thank you!!

PR #172, opened 2020-02-20
Update roles pages to match mocks

karelhala:
Screenshot from 2020-02-20 10-47-26
Screenshot from 2020-02-20 10-47-34
Screenshot from 2020-02-20 10-47-44

### DIFF
--- a/src/smart-components/role/role-permissions.js
+++ b/src/smart-components/role/role-permissions.js
@@ -74,6 +74,7 @@ const Permissions = () => {
       } }
       titlePlural="permissions"
       titleSingular="permission"
+      filterPlaceholder="Filter by application"
     />
   </section>;
 };


karelhala:
ping @mmenestr

mmenestr:
Looks good to me! Is this the PR for this story about filters? https://projects.engineering.redhat.com/browse/RHCLOUD-4257 or a different one?

Also, I hope this isn't too much hassle, but I actually liked the fact that your initial design of the permissions filter had "Filter by permission" since the user could search by any column (not just the application column). Would it be possible to change it to an attribute filter to filter by all the columns OR could you change it back to "Filter by permission" as a basic filter.

karelhala:
@mmenestr I found out that the roles filter doesn't actually filter in detail. I am writing new PR to fix that, if you want I can include the previous filtering as well.

mmenestr:
So you found out that for roles in groups, you can now filter by name and description:
**Takeaway:** 
1. For Roles in group change filter to an attribute filter (attribute 1: Name, attribute 2: Description)
2. For permissions, change back to filtering by any field (Search field should say: "Filter by permission")

PR #175, opened 2020-02-20
Add filter for roles in group

karelhala:
### Filtering by multiple fields
Screenshot from 2020-02-20 16-09-20
### Revert of filtering in permissions
Screenshot from 2020-02-20 16-10-07


karelhala:
@mmenestr this is the updated filter, based on your suggestions from https://github.com/RedHatInsights/insights-rbac-ui/pull/172#issuecomment-589087133

mmenestr:
Last thing - the name attribute is the one that shows up by default right? like if the user didn't touch the filter at all, and is just coming onto the page for the first time, they would see "Name" as the attribute, and would have to click into it to then see "Description" correct?

karelhala:
@mmenestr yup, this is how it looks like
Screenshot from 2020-02-20 16-45-38

mmenestr:
Great! It's good to go then, thank you!!

PR #176, opened 2020-02-20
Use frontend filtering and pagination for members in group

karelhala:


src/smart-components/group/principal/principals.js
```diff
@@ -102,6 +112,7 @@ const GroupPrincipals = ({
             variant: 'danger'
           },
           onClick: () => {
+            console.log(selectedPrincipals);
```
Hyperkid123: :dagger:

PR #181, opened 2020-02-26
Apply sorting to groups table

fhlavac:
- added sorting to table-toolbar-view
- applied sorting to groups table
- tested fetching groups on sorting button click

@karelhala

### DIFF
--- a/src/presentational-components/shared/table-toolbar-view.js
+++ b/src/presentational-components/shared/table-toolbar-view.js
@@ -1,4 +1,4 @@
-import React, { Fragment, useState } from 'react';
+import React, { Fragment, useState, useEffect } from 'react';
 import propTypes from 'prop-types';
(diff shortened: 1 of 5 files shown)


src/presentational-components/shared/table-toolbar-view.js
```diff
@@ -113,6 +121,18 @@ export const TableToolbarView = ({
         { ...rows.length > 0 && { actionResolver } }
         areActionsDisabled={ areActionsDisabled }
         rowWrapper={ rowWrapper }
+        sortBy={ sortByState }
+        onSort={ (e, index, direction) => {
+          const key = columns[index - isSelectable].key;
```
karelhala: No need for this, it can be simplified
```JS
setSortByState({ index, direction });
fetchData({
            ...pagination,
            offset: 0,
            name: '',
            orderBy: `${direction === 'desc' ? '-' : ''}${columns[index - isSelectable].key}`
          });
```

PR #185, opened 2020-02-27
Update members to use new API and filter by username

karelhala:


src/helpers/group/group-helper.js
```diff
@@ -60,3 +60,12 @@ export async function addRolesToGroup(groupId, roles) {
   return await groupApi.addRoleToGroup(groupId, { roles });
 }
 
+export async function fetchPrincipalsForGroup(groupId, usernames, pagination, options = {}) {
+  return await groupApi.getPrincipalsFromGroup(groupId, usernames, {
+    ...options || {},
+    query: {
+      ...options.query || {},
```
Hyperkid123: I don't think you need the `|| {}` part. `...` works on undefined

Hyperkid123: I don't think you need the `|| {}` part. `...` works on undefined

package.json
```diff
@@ -65,7 +65,7 @@
     "babel-plugin-lodash": "^3.3.4",
     "clean-webpack-plugin": "^0.1.19",
     "connect-history-api-fallback": "^1.5.0",
-    "copy-webpack-plugin": "^4.5.1",
+    "copy-webpack-plugin": "^5.1.1",
```
rvsia: This plugin is probably useless, as it just copies the old Red Hat logo. :smile_cat:

rvsia: This plugin is probably useless, as it just copies the old Red Hat logo. :smile_cat:

rvsia: This plugin is probably useless, as it just copies the old Red Hat logo. :smile_cat:

rvsia: This plugin is probably useless, as it just copies the old Red Hat logo. :smile_cat:

PR #186, opened 2020-02-27
Fix group roles filter

PanSpagetka:
Fix filtering Group roles with multiple different filters


mmenestr:
Are there any screenshots I can see?

karelhala:
@mmenestr not really, this was a bug that when we applied name filter and description filter on the one that was changed as last was applied. Basically the chips were lying.

PR #204, opened 2020-03-09
Enable deleting custom roles

PanSpagetka:
For some reason, deleting roles is disabled. It shows the three dot action menu only for custom roles, but with disabled button. I am enabling it with this PR.

Is there any reason why it should stay disabled?
@karelhala @ryelo @shdunning @nicole-campbell12

Screenshot from 2020-03-09 13-50-13

When i delete role, UI blows up, but role is deleted. I will look into it further if we decide that this button should be enabled.

Also Add role and Confirm delete modals looks like in temporary version when compared to other parts of RBAC UI. Is there a plan to redesign it?  @mmenestr

Screenshot from 2020-03-09 13-56-11

cc @john-dupuy

### DIFF
--- a/src/smart-components/role/roles.js
+++ b/src/smart-components/role/roles.js
@@ -57,11 +57,7 @@ const Roles = () => {
       {
         title: 'Delete',
         onClick: (_event, _rowId, role) =>
-          push(`/roles/remove/${role.uuid}`),
-        props: {
-          isDisabled: true
-        },
-        isDisabled: true
+          push(`/roles/remove/${role.uuid}`)
       }
     ];
   };
--- a/src/smart-components/role/roles.js
+++ b/src/smart-components/role/roles.js
@@ -52,7 +52,7 @@ const Roles = () => {
   </Fragment>;
 
   const actionResolver = ({ system }) => {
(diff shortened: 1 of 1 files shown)


mmenestr:
@PanSpagetka The way it is now is correct (don't change it haha)!! Users can not delete a role from the system unless they've created it themselves. They can only delete a role if it's inside a group. 
So again, users should NOT be able to delete a "canned" role from the roles page.

mmenestr:
Also yes, I plan on redesigning these soon, thanks for pointing the modal out too! 
Since barely anyone will have the capability of creating or deleting a role, it was not prioritized for this initial RBAC launch, and it'll be addressed after the launch :)

nicole-campbell12:
@PanSpagetka any org admin should be able to delete a custom role; only pre-defined (aka canned) roles created by platform should be immutable.

PanSpagetka:
@nicole-campbell12 @mmenestr Thank you, that is what this PR should do.

I will look further into UI blow up after deleting role, marking this PR as WIP until it is fixed

PanSpagetka:
It should be working now :)

PR #205, opened 2020-03-09
Re-enable role filters in group detail

karelhala:


karelhala:
@PanSpagetka is there any reason why the filter was used there? This actually broke showing the filters.

PR #207, opened 2020-03-10
Enable selecting steps in role creation wizard

PanSpagetka:
I have fixed the logic that enabled steps after completing. So you can navigate through completed steps but you cant skip steps. You can navigate between valid steps. That means if you fill the whole wizard, then go to the first step, delete required field, navigation becomes disabled until you fill required field. This might be a bit confusing, so I have recorded video.

https://projects.engineering.redhat.com/browse/RHCLOUD-4893

#### Video
https://webmshare.com/play/Kx6K1

@john-dupuy @nicole-campbell12


john-dupuy:
Looks great, thanks for attaching the video @PanSpagetka

src/smart-components/role/add-role/role-information.js
```diff
@@ -42,7 +42,7 @@ const RoleInformation = (formData, onHandleChange) => {
               id="description"
               name="description"
               value={ description }
-              onChange={ (_, event) => onHandleChange({ description: event.currentTarget.value }) }
+              onChange={ (_, event) => onHandleChange({ description: event.currentTarget.value }, true) }
```
karelhala: Shouldn't this be `event.currentTarget.value.trim().length > 0`?

PanSpagetka: If you mean the second parameter like: `onHandleChange({ description: event.currentTarget.value, event.currentTarget.value.trim().length > 0 }`  than no. Description is not required, so the form is valid even if user doesn't write anything there.

karelhala: Oh, I see. Okay, then it's good to go!

PR #254, opened 2020-04-06
Fix group-members pagination

fhlavac:
* fixed not updating count when member added
* fixed page not changing change on pagination arrow click

@karelhala

### DIFF
--- a/src/redux/reducers/group-reducer.js
+++ b/src/redux/reducers/group-reducer.js
@@ -6,7 +6,7 @@ import {
   FETCH_ADD_ROLES_FOR_GROUP,
   FETCH_MEMBERS_FOR_GROUP
 } from '../../redux/action-types';
-import _ from 'lodash';
+import omit from 'lodash/omit';
 
 // Initial State
 export const groupsInitialState = {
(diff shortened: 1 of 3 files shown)


src/redux/reducers/group-reducer.js
```diff
@@ -6,6 +6,7 @@ import {
   FETCH_ADD_ROLES_FOR_GROUP,
   FETCH_MEMBERS_FOR_GROUP
 } from '../../redux/action-types';
+import _ from 'lodash';
```
karelhala, line 9:
```suggestion
import omit from 'lodash/omit';
```

src/redux/reducers/group-reducer.js
```diff
@@ -34,7 +35,7 @@ const setGroup = (state, { payload }) => ({
       ...payload.uuid === group.uuid && { ...payload, loaded: true }
     }))
   },
-  selectedGroup: { ...state.selectedGroup, members: { ...state.selectedGroup.members, data: payload.principals }, ...payload, loaded: true }
+  selectedGroup: { ...state.selectedGroup, members: { ...state.selectedGroup.members, data: payload.principals }, ..._.omit(payload, 'principals'), loaded: true }
```
karelhala, line 38:
```suggestion
  selectedGroup: { ...state.selectedGroup, members: { ...state.selectedGroup.members, data: payload.principals }, ...omit(payload, 'principals'), loaded: true }
```

PR #271, opened 2020-04-16
Add info about inactive users

karelhala:
Screenshot from 2020-04-17 10-36-07
Screenshot from 2020-04-17 10-36-37
Screenshot from 2020-04-17 10-37-01

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -102,3 +102,13 @@
     }
   }
 }
+
+.ins-c-rbac__user-label {
+  margin-right: var(--pf-global--spacer--sm);
+
+  &.ins-m-inactive { background-color: var(--pf-global--disabled-color--100); }
+}
+
+.ins-c-rbac__user-row {
+  &.ins-m-inactive { color: var(--pf-global--disabled-color--200); }
+}
--- a/src/App.scss
+++ b/src/App.scss
@@ -91,8 +91,12 @@
   color: var(--pf-global--warning-color--100);
 }
 
-.pf-c-table tbody tr.ins-c-rbac__group-default {
-  .pf-c-table__check:first-of-type input { display: none; }
+.pf-c-table {
+  thead tr th.ins-m-width-5 { width: 5%; }
+
+  tbody tr.ins-c-rbac__group-default {
+    .pf-c-table__check:first-of-type input { display: none; }
+  }
 }
 
 .ins-c-rbac__summary {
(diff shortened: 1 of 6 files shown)


karelhala:
ping @mmenestr

john-dupuy:
@karelhala I have some concerns that this might break the widget we use to represent a `Table` in c.rh.c. When we lookup a user in the table we match against the username text. I would have to check to be sure, but now the username text might also include that "Active/Inactive" text before the username. 

Apologies if this was discussed on ticket @mmenestr, but is there any issue UX wise if the chip/label is in a separate (untitled) column? If so, I'm sure I can figure out a workaround.

karelhala:
@john-dupuy yeah, you are right, now part of username is Active/Inactive label. Mhh I could move the label to specific column before the username if that would help.

mmenestr:
No issue with moving the label to another separate column! I would prefer that based on what you're saying :) 

Also 2 other notes I have:
1. **The description in the user page should be "These are all the users in your Red Hat organization. To add new users or manage existing users, go to your user management list."** It's just a little longer than the ones in the modals.

2. The disabled text color seems a little too light? But maybe it's the screenshot. Could you use --pf-global--disabled-color--100 if not already?

karelhala:
> No issue with moving the label to another separate column! I would prefer that based on what you're saying :)
> 
> Also 2 other notes I have:
> 
> 1. **The description in the user page should be "These are all the users in your Red Hat organization. To add new users or manage existing users, go to your user management list."** It's just a little longer than the ones in the modals.

Sure thing!

> 2. The disabled text color seems a little too light? But maybe it's the screenshot. Could you use --pf-global--disabled-color--100 if not already?

I've used `pf-global--disabled-color--100` for the inactive label and `pf-global--disabled-color--200` for the color of inactive row.

john-dupuy:
Yup, in a separate column there'd be no problem for our widget at all!

mmenestr:
The color of the row text is too light! Could you use --pf-global--disabled-color--100 instead so that the username, name etc are more readable?

karelhala:
Oh you mean the color of the text in a row, sure thing! I'll update it to use `-100`.

karelhala:
This is how it looks like with lighter color
Screenshot from 2020-04-17 10-15-45

karelhala:
@mmenestr screenshots updated. I've added table cell header for the Active/Inactive part so users know what it means.

mmenestr:
Ok also LGTM! Thanks Karel!

PR #278, opened 2020-04-20
Change default pagination in all tables to 20

fhlavac:
- default pagination limit in all tables is now 20
- in modal tables, the default limit is 10
- updated tests

@karelhala

### DIFF
--- a/src/helpers/shared/pagination.js
+++ b/src/helpers/shared/pagination.js
@@ -1,11 +1,13 @@
 export const defaultSettings = {
   limit: 20,
-  offset: 0
+  offset: 0,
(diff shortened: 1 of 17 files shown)


src/helpers/shared/pagination.js
```diff
@@ -1,16 +1,11 @@
 export const defaultSettings = {
-  limit: 50,
-  offset: 0,
-  itemCount: 1,
```
karelhala, line 4: I think that the default settings should have `itemCount` at least set to 0 so if the server reposnds with something strange we won't break pagination.

PR #282, opened 2020-04-22
Add User detail

PanSpagetka:
Add User detail

Screenshot from 2020-04-27 12-10-02
Screenshot from 2020-04-27 12-09-49
Screenshot from 2020-04-27 12-10-02


src/smart-components/user/user.js
```diff
@@ -0,0 +1,229 @@
...
+                { title: `${groups_in.length}`, props: { isOpen: expanded[uuid] === 1 }},
+                { title: accessCount, props: { isOpen: expanded[uuid] === 2 }},
+                { title: <DateFormat type='exact' date={ modified } /> }
+                ]
+            }, {
+                uuid: `${uuid}-groups`,
+                parent: 3 * i,
+                compoundParent: 1,
+                cells: [{
+                    title: <Table
```
karelhala, line 68: You should add props to make the inner table full width `props: { colSpan: 2, className: 'pf-m-no-padding' }`

src/smart-components/user/user.js
```diff
@@ -0,0 +1,229 @@
...
+                        <TableBody />
+                    </Table>
+                }]
+            },
+            {
+                uuid: `${uuid}-groups`,
+                parent: 3 * i,
+                compoundParent: 2,
+                cells: [{
+                    title: !isRecordLoading && rolesWithAccess && rolesWithAccess[uuid]
```
karelhala, line 85: Same as above `props: { colSpan: 3, className: 'pf-m-no-padding' }`

src/smart-components/user/user.js
```diff
@@ -0,0 +1,229 @@
...
+                        }
+                    }]} }
+                    pagination={ { itemCount: roles.meta.count, page, perPage, onSetPage, onPerPageSelect } }
+                />
+                { isLoading
+                ? <ListLoader />
+                : <Table
+                    className='ins-c-activity-table'
+                    aria-label="Collapsible table"
+                    onExpand={ onExpand }
```
karelhala, line 182: I'd extend the `TableToolbarView` to consume `onExpand` function this way we can easilly use this table variant without the need to copy this code all over the places.

src/smart-components/user/user.js
```diff
@@ -0,0 +1,229 @@
...
+    return {
+       roles,
+       isLoading,
+       rolesWithAccess,
+       isRecordLoading
+    };
+};
+
+const mapDispatchToProps = dispatch => {
+    return {
```
karelhala, line 223: No need to have return and block, you can directly expose the object, I guess this should use same default pagination as other tables.
```JS
const mapDispatchToProps = dispatch => ({
  fetchRoles: (apiProps = { limit: 10, offset: 0 }) => dispatch(fetchRoles(apiProps)),
  fetchRoleForUser: (uuid) => dispatch(fetchRoleForUser(uuid))
});
```

src/smart-components/user/user.js
```diff
@@ -0,0 +1,229 @@
...
+    isLoading: PropTypes.bool,
+    rolesWithAccess: PropTypes.object,
+    isRecordLoading: PropTypes.bool
+};
+
+User.defaultProps = {
+
+};
+
+const mapStateToProps = ({ roleReducer: { roles, isLoading, rolesWithAccess, isRecordLoading }}) => {
```
karelhala, line 212: No need to have block and return here
```JS
const mapStateToProps = ({ roleReducer: { roles, isLoading, rolesWithAccess, isRecordLoading }}) => ({
  roles,
  isLoading,
  rolesWithAccess,
  isRecordLoading
});
```

src/smart-components/user/user.js
```diff
@@ -0,0 +1,229 @@
...
+    match: PropTypes.object,
+    fetchRoles: PropTypes.func,
+    fetchRoleForUser: PropTypes.func,
+    roles: PropTypes.array,
+    isLoading: PropTypes.bool,
+    rolesWithAccess: PropTypes.object,
+    isRecordLoading: PropTypes.bool
+};
+
+User.defaultProps = {
```
karelhala, line 208: This can be removed.

src/smart-components/group/add-group/users-list.js
```diff
@@ -31,7 +33,7 @@ const createRows = (data, expanded, checkedRows = []) => {
         props: {
           data: { isActive }
         }
-      }, username, email, first_name, last_name ],
+      }, { title: <Link to={ `/users/detail/${username}` }>{username}</Link> }, email, first_name, last_name ],
```
karelhala, line 36: This will add the link in all places where users-list is used (when adding users to group etc.) I don't think we want that. I'd hide this link behind some prop.

src/redux/reducers/role-reducer.js
```diff
@@ -19,10 +20,17 @@ const setLoadingState = state => ({ ...state, isLoading: true });
 const setRecordLoadingState = state => ({ ...state, isRecordLoading: true });
 const setRole = (state, { payload }) => ({ ...state, selectedRole: payload, isRecordLoading: false });
 const setRoles = (state, { payload }) => ({ ...state, roles: payload, isLoading: false });
+const setRolesWithAccess = (state, { payload }) => ({
+  ...state,
+  rolesWithAccess: { ...state.rolesWithAccess, [payload.uuid]: payload },
```
karelhala, line 25: You should add the `isLoading` state to `rolesWithAccess` this way you can have more than one row loading at a time.
```JS
{
  rolesWithAccess: {
     [uuid]: {
       isLoadig: true,
       data: payload
     }
  }
}
```

PanSpagetka: Loading multiple rows seems to work fine. To do this I would have do set loading for each uuid in pending action. But in pending action I don't get uuid, only the type. I could probably overcome it somehow, but I don't think it is necessary. Because I am accessing "random" key I am checking for existence of the value anyway.

mmenestr:
This all looks correct, thank you!! Once you push, tell me so I can double check the linking and stuff!

The only comment I have is that it looks really weird under permissions, that only half the expanded table is populated with information, I wonder if it would be possible to center that information? or make the grey lines expand till the end of the box at least? 

Screen Shot 2020-04-24 at 8 28 53 AM

mmenestr:
Update: so it seems like all the content is only scoped to the first column, when it should actually span the whole width of the table, and thats why all the data is pushed to just the left side. Could you fix it so that it spans the whole area, and the columns are equally spread out? @ryelo don't know if you have any more technical help to give here

Also, that applies to both the groups and the permissions expanded tables! thank you!!

src/smart-components/user/user.js
```diff
@@ -0,0 +1,222 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import AwesomeDebouncePromise from 'awesome-debounce-promise';
```
karelhala, line 4: Is this really neccessary? We've been using lodash debounce for a while

src/smart-components/user/user.js
karelhala: Actually, I've checked and it should have colSpan of parent table, so here it would be 4.

src/smart-components/user/user.js
karelhala: Actually, I've checked and it should have colSpan of parent table, so here it would be 4.

PanSpagetka:
@karelhala @mmenestr  have updated screenshots. Hopefully all coding and visual issues are fixed.

package.json
```diff
@@ -3,17 +3,18 @@
   "version": "1.1.0",
   "private": false,
   "dependencies": {
+    "awesome-debounce-promise": "2.1.0",
```
karelhala, line 6: You don't really need this installed, right?

src/smart-components/user/user.js
```diff
@@ -0,0 +1,199 @@
...
+            setRows(createRows(roles.data));
+        }
+    }, [ roles ]);
+
+    useEffect(() => {
+        setRows(createRows(roles.data));
+    }, [ rolesWithAccess ]);
+
+    const onExpand = (event, rowIndex, colIndex, isOpen, rowData) => {
+        const r = [ ...rows ];
```
karelhala, line 122: I don't think you'd have to copy the rows in here. You can just use `rows`.

PanSpagetka: Actually I don't need it at all :laughing:

karelhala:
@PanSpagetka let's wait for @mmenestr to give us approve to merge it. Overall looks good.

mmenestr:
Looks great! Thank you for waiting for me to look at it :) Is it going to be possible to get the group description in the end? I see that the header is there but the info is not!

PanSpagetka:
It will eventually be there when API returns it

coderbydesign:
@mmenestr @PanSpagetka the description PR was merged https://github.com/RedHatInsights/insights-rbac/pull/279 and deployed to CI.

PR #289, opened 2020-04-29
Move to first page after group member/role delete and reload data

fhlavac:
* after deleting group role/member push to the first page
* after adding/removing group role reload data

@karelhala

### DIFF
--- a/src/smart-components/group/principal/principals.js
+++ b/src/smart-components/group/principal/principals.js
@@ -64,7 +64,10 @@ const GroupPrincipals = () => {
   };
 
   const removeMembers = (userNames) => {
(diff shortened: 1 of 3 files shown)


src/smart-components/group/principal/principals.js
```diff
@@ -64,7 +64,7 @@ const GroupPrincipals = () => {
   };
 
   const removeMembers = (userNames) => {
-    return dispatch(removeMembersFromGroup(uuid, userNames)).then(() => { setSelectedPrincipals([]); fetchData();});
+    return dispatch(removeMembersFromGroup(uuid, userNames)).then(() => { setSelectedPrincipals([]); fetchData(undefined, { ...pagination, offset: 0 });});
```
karelhala, line 67: Can you style this row so it'll be on 2 lines? It will be easier to read.

PR #331, opened 2020-06-17
Enhance group roles fetch data and snapshot tests

fhlavac:
- enhanced snapshot test
- added fetch data test

@karelhala


src/smart-components/group/role/group-roles.js
```diff
@@ -58,6 +58,7 @@ const addRoleButton = (isDisabled) => (isDisabled
     </div>
   </Tooltip>
   : <Button
+    id="add-role-btn"
```
karelhala, line 61: Is adding ID really necessary? I'd like to avoid as much as possible, if you want to find element in shadow dom while testing you can use `at` method.

PR #333, opened 2020-06-23
Update DDF to v2

fhlavac:
- updated DDF to version 2.4.8
- fixed edit-group-modal
- fixed tests

00ě

@karelhala

PR #335, opened 2020-06-24
Custom role Wizard first page

PanSpagetka:
#333 should be merged first

https://projects.engineering.redhat.com/browse/RHCLOUD-7235
Screenshot from 2020-06-29 14-33-22
Screenshot from 2020-06-29 14-33-03


src/presentational-components/shared/toolbar.js
```diff
@@ -199,9 +199,6 @@ const Toolbar = ({
     { ...!isLoading && {
       pagination: paginationBuilder(pagination, fetchData, filterValue, sortBy)
     } }
-    { ...(filterValue.length > 0 || (textFilters && textFilters.length > 0)) && {
```
karelhala, line 202: Is there a reason to remove this? Cause this generates active filter chips.

src/smart-components/role/add-role-new/schema.js
```diff
@@ -0,0 +1,134 @@
+// export const asyncValidator = async (value, sourceId = undefined) => {
```
karelhala, line 1: No need to have comments.

mmenestr:
Looks good to me from a visual standpoint! Will look again once it's merged for interaction standpoint :)

mmenestr:
@PanSpagetka  Hi, just went through it and noticed one small thing that needs changing. 
See recording here:
[Screen Recording 2020-07-01 at 9.44.45 AM.zip](https://github.com/RedHatInsights/insights-rbac-ui/files/4858254/Screen.Recording.2020-07-01.at.9.44.45.AM.zip)

But basically, if I click the option "Copy an existing role" and make a selection, the "Next" button turns blue. Which is correct.
But if I do that, and THEN change my mind and say "Create a role from scratch" and then go back to "Copy an existing role", the role has been unselected, but the "Next" button remains blue. Since nothing is selected though, the user shouldn't be able to press the "Next" button anymore. This could be fixed in 2 ways, either:

(1) have the selection in the table remain, even as you switch between the main radio buttons
OR
(2) Have the selection de-select like it does now, but make sure the button stays disabled if there is no selection in the table!

PanSpagetka:
@mmenestr Yep, it seems to be clearing the selection. @fhlavac will take a look ;)

fhlavac:
@mmenestr this should fix it https://github.com/RedHatInsights/insights-rbac-ui/pull/343

PR #341, opened 2020-07-01
Fix activity label initial state in user detail

fhlavac:
https://projects.engineering.redhat.com/browse/RHCLOUD-7507

Active state should not be changing during the load anymore.
The problem was that wrong user record used to be loaded from the redux state initially and after fetch it was rewritten by the correct one. Now, the correct user is passed from the previous state.

@john-dupuy 
@karelhala

### DIFF
--- a/src/smart-components/user/user.js
+++ b/src/smart-components/user/user.js
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


src/smart-components/user/user.js
```diff
@@ -179,11 +178,11 @@ User.propTypes = {
 const mapStateToProps = ({
     roleReducer: { roles, isLoading, rolesWithAccess },
     userReducer: { users: { data }}
-}) => ({
+}, username) => ({
```
karelhala, line 181: I think that this should be `match: { params: { username }}`

PR #342, opened 2020-07-07
Add permissions step

PanSpagetka:
Add permissions step to custom role wizard. So far using `/access` endpoint, however eventually it will use different endpoint here: https://projects.engineering.redhat.com/browse/RHCLOUD-6573 

So far pagination and filtering is client side.

https://projects.engineering.redhat.com/browse/RHCLOUD-7258

Screenshot from 2020-07-08 13-46-53


src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -0,0 +1,200 @@
...
+            row.selected = selected;
+            return row;
+        }));
+    };
+
+    return (isLoading
+        ? 'LOADING'
+        : <div>
+
+            <PrimaryToolbar
```
karelhala, line 116: Can we use `TableToolbarView`? It will simplify a lot of boilerplate code.

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -0,0 +1,200 @@
...
+import { getPrincipalAccess } from '../../../redux/actions/access-actions';
+import useFieldApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-field-api';
+
+const columns = [ 'Application', 'Resource type', 'Operation' ];
+const selector = ({ accessReducer: { access, isLoading }}) => ({
+    access: access.data,
+    isLoading
+});
+const types = [ 'application', 'resource', 'operation' ];
+const accessWrapper = (rawData) => (filters = []) => {
```
karelhala, line 20: Phew this looks interesting, you should write some tests for it so we don't break anything crucial in future.

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -0,0 +1,115 @@
...
+                typeof operations !== 'undefined' && setFilters({ ...filters, operations });
+            } }
+            isLoading={ isLoading }
+            pagination={ { ...pagination, count: permissions.filteredData.length } }
+            checkedRows={ selectedPermissions }
+            setCheckedItems={ setCheckedItems }
+            titlePlural="permissions"
+            titleSingular="permission"
+            hideFilterChips
+            textFilters={ [
```
karelhala, line 95: Good for now, but we might want to reconsider changing this prop to `filters` instead of `textFilters`

PanSpagetka: Yep, I agree. But I would do it in separate PR.

PR #363, opened 2020-07-14
Remove "[object Object]" from userlist DOM

fhlavac:
removed "[object Object]" from DOM of the status column in the Users table

 https://projects.engineering.redhat.com/browse/RHCLOUD-7741

@karelhala is the `data` value "Active"/"Inactive" ok?

@john-dupuy

### DIFF
--- a/src/presentational-components/shared/UsersRow.js
+++ b/src/presentational-components/shared/UsersRow.js
@@ -1,11 +1,13 @@
 import React from 'react';
 import { RowWrapper } from '@patternfly/react-table';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 const UsersRow = ({ row, ...props }) => {
   const { status } = row;
+  const { 'data-is-active': isActive } = status?.props;
   return <RowWrapper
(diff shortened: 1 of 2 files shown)


karelhala:
> @karelhala is the data value "Active"/"Inactive" ok?

Good call, AFAIK this value is used in [src/presentational-components/shared/UsersRow.js#L8](https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/presentational-components/shared/UsersRow.js#L8), so either change this line to `status?.props?.data` but I'd change the stored value completely from `props: { data: { isActive } }` to `props: { datum: { isActive } }`

fhlavac:
> > @karelhala is the data value "Active"/"Inactive" ok?
> 
> Good call, AFAIK this value is used in [src/presentational-components/shared/UsersRow.js#L8](https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/presentational-components/shared/UsersRow.js#L8), so either change this line to `status?.props?.data` but I'd change the stored value completely from `props: { data: { isActive } }` to `props: { datum: { isActive } }`

updated the attribute to `data-is-active` as we agreed, thanks @karelhala!

PR #376, opened 2020-07-20
Add name validation for Custom Roles

PanSpagetka:
https://projects.engineering.redhat.com/browse/RHCLOUD-7465
Screenshot from 2020-07-20 14-02-26
Screenshot from 2020-07-20 14-13-08
Screenshot from 2020-07-20 14-11-49
Screenshot from 2020-07-20 14-12-02

### DIFF
--- a/src/smart-components/role/add-role-new/validators.js
+++ b/src/smart-components/role/add-role-new/validators.js
@@ -1,0 +1,24 @@
+import { fetchRoles } from '../../../helpers/role/role-helper';
+import asyncDebounce from '../../../utilities/async-debounce';
+
+const asyncValidator = async (groupName) => {
+    if (!groupName) {
+        return undefined;
+    }
+
+    let response;
+    try {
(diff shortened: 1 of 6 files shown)


src/smart-components/role/add-role-new/validators.js
```diff
@@ -0,0 +1,24 @@
+import { fetchRoles } from '../../../helpers/role/role-helper';
+import asyncDebounce from '../../../utilities/async-debounce';
+
+const asyncValidator = async (groupName) => {
+    if (!groupName) {
+        return undefined;
+    }
+
+    let response;
```
karelhala, line 9: You can catch directly on the response
```JS
const response = await fetchRoles({ limit: 10, offset: 0, name: groupName, nameMatch: 'exact' })
.catch(e => showNotificationError());

if (response?.data?.length > 0) {
        throw 'Name has already been taken';
    }
```

PR #393, opened 2020-07-30
Add group name validation

fhlavac:
https://projects.engineering.redhat.com/browse/RHCLOUD-7236

- added group name duplicate validation to group information step
- added error page when group adding fails because of an existing group with the same name
- updated tests

Thanks for help! @karelhala @PanSpagetka


src/smart-components/group/add-group/add-group-wizard.js
```diff
@@ -57,9 +60,13 @@ const AddGroupWizard = ({
...
     };
-    await addGroup(user_data);
-    postMethod();
-    history.push(closeUrl);
+    try {
+      await addGroup(user_data);
+      postMethod();
+      history.push(closeUrl);
+    } catch (e) {
+      setIsGroupInfoValid(false);
```
PanSpagetka, line 68: Does api return what kind of error happened? Can we differentiate the name is taken error and other errors?

PR #395, opened 2020-07-31
Add sitemap generator

PanSpagetka:
This is example how can sitemap be generator for application.
First you need to add `react-router-sitemap` to `package.json`
Second you need to add routes into `package.json`.
Then you can generate sitemap by executing `npm run sitemap`.
You can also use routes defined in `packege.json` in you application code. `import { routes } from '../../../package.json';`


ryelo:
@PanSpagetka When this gets the green check, do you mind adding all of this to the starter app too? We'll point teams to the commit you'll add there just to make it super simple to follow.

src/sitemap-builder.mjs
```diff
@@ -0,0 +1,12 @@
+import Sitemap from 'react-router-sitemap';
+import fs from 'fs';
+const jsonString = fs.readFileSync('/home/panspagetka/insights/insights-rbac-ui/package.json');
```
karelhala, line 3: This will not work, can you change it to `import { routes } from './package.json'`?

src/sitemap-builder.mjs
```diff
@@ -0,0 +1,12 @@
...
+const { routes } = JSON.parse(jsonString);
+
+const paths = Object.entries(routes).map(([ _, value ]) => value.path || value);
+
+const hostname = 'http://cloud.redhat.com';
```
karelhala, line 8: Shouldn't this be `https`?

src/sitemap-builder.mjs
```diff
@@ -0,0 +1,12 @@
+import Sitemap from 'react-router-sitemap';
+import fs from 'fs';
+const jsonString = fs.readFileSync('/home/panspagetka/insights/insights-rbac-ui/package.json');
+const { routes } = JSON.parse(jsonString);
+
+const paths = Object.entries(routes).map(([ _, value ]) => value.path || value);
```
karelhala, line 6: If you are not going to use the key, it's better to use `Object.values`

src/sitemap-builder.mjs
PanSpagetka: Thanks for the catch ;)  Node doesn't like `import { routes } from './package.json'` but `import packageJson from './package.json'` is just fine.

sitemap.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:mobile="http://www.google.com/schemas/sitemap-mobile/1.0" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1" xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">
+<url> <loc>http://cloud.redhat.com/</loc> </url>
```
karelhala, line 3: Wondering why it's still http 🤔

PanSpagetka: That's because I didn't include didn't generate new `sitemap.xml`

karelhala:
Looking hella good! If you want to learn about plugins for webpack, I'd make this into a plugin and apps would just have to include list of paths in `package.json` and that would be it!

PR #398, opened 2020-08-05
fix the denied state

ryelo:


src/App.js
```diff
@@ -43,7 +43,7 @@ class App extends Component {
       return <AppPlaceholder />;
     }
 
-    if (!isAdmin && insights.chrome.getApp() === 'rbac') {
+    if (!isAdmin) {
```
PanSpagetka, line 46: Won't this cause that `My user access` page will be also denied?

ryelo: MUA is should only be available in ci-beta and it's going through a redesign anyway

PR #400, opened 2020-08-06
Custom Roles - use permission text filter with "See more"

fhlavac:
https://projects.engineering.redhat.com/browse/RHCLOUD-8436

- updated PF versions
- added conditional filter with "See more" to Custom roles - add permissions
- updated snapshots

uses https://github.com/RedHatInsights/frontend-components/pull/704

@karelhala 
@PanSpagetka 
@mmenestr


mmenestr:
LGTM! And you're going to add a search in the menu dropdown too right?

karelhala:
> LGTM! And you're going to add a search in the menu dropdown too right?

I don't think it's a good idea to have it there. Once user types in there the list is limited to let's say 5 items and once user clicks outside of the dropmenu they don't know what they inserted into the filter. From user experience I would use the in menu filter only if there is finite amount of things without show more/less button and there are no async calls.

What I usually do with these kind of menus is that I open them, type something in there to initiate the search, when I see the spinner I often close the menu, just to look at the rest of the UI (no need to look at spinning icon) and after some time I click on it to see if the async action finished. If the filter would be only in the menu I'd freak out and clicked again just to see if the text actually stayed there. I would then be suspicious if the filter actually got applied. Not to speak about the use case, that I would have searched for something, clicked away because I wouldn't get what I actually wanted scrolled down to see if the thing was actually in the table and check that the spelling is correct in the search, that would require additional clicking on the search. I really think that this would deserve some user study with different usesaces and behaviors, or even A/B testing.

mmenestr:
Oh wait, can the user search in the toggle where it says "Filter by application"? Like is it a type ahead. If it's already a type ahead then no need to add a second search

karelhala:
@mmenestr yup, that is the autocomplete component, the `Filter by application` is just a placeholder and users can type in there what they want and it would filter the applications.

fhlavac:
@karelhala @PanSpagetka 
moved the filtering & showMore logic to RBAC, hope it's better now

karelhala (commented):
It looks really good from code side! I'll give it a try locally if it doesn't break any other filter and after that merge it.

PanSpagetka (changes requested):
I have tested in Custom roles wizard and it seems that there is an issue.
When you select a few applications then you select some resources, when you return to to applications, no application is selected in the menu (it is in the chips)  and when you select something it clears you previous selection.

PanSpagetka (approved):
It seems to work :)

karelhala (approved):

PR #420, opened 2020-08-20
Fix selected group loading

PanSpagetka:
https://projects.engineering.redhat.com/browse/RHCLOUD-7546


john-dupuy:
@PanSpagetka when I run this PR locally, I am still able to see that the title of the first group displays while loading the second. Is there some way we can force a loading state here?

PanSpagetka:
@john-dupuy hmm, when I am testing it, it seems that there are always loading bars. But maybe I am missing something...I will get in touch with you

karelhala (approved):

PR #421, opened 2020-08-25
Resolve splats fix

PanSpagetka:
Fix not resolving splats from base role. Add tests for `resolveSpalts` function.

https://projects.engineering.redhat.com/browse/RHCLOUD-7532

@john-dupuy

### DIFF
--- a/src/smart-components/role/add-role-new/add-permissions.js
+++ b/src/smart-components/role/add-role-new/add-permissions.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { shallowEqual, useSelector, useDispatch } from 'react-redux';
 import useFieldApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-field-api';
 import useFormApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-form-api';
+import flatMap from 'lodash/flatMap';
 
 import { TableToolbarView } from '../../../presentational-components/shared/table-toolbar-view';
 import { listPermissions } from '../../../redux/actions/permission-action';
@@ -32,12 +33,29 @@ export const accessWrapper = (rawData, filters = { applications: [], resources: 
     );
 
     return {
-        data,
+        data: data.filter(permission => filterSplats(permission)),
         filteredData,
         applications: [ ...new Set(data.filter(permission => filterResource(permission) && filterOperation(permission)).map(({ application }) => application)) ],
         resources: [ ...new Set(data.filter(permission => filterApplication(permission) && filterOperation(permission)).map(({ resource }) => resource)) ],
         operations: [ ...new Set(data.filter(permission => filterApplication(permission) && filterResource(permission)).map(({ operation }) => operation)) ]
     };
+};
+
+export const resolveSplats = (selectedPermissions, permissions) => {
+    return (permissions.length > 0
+        ? flatMap(selectedPermissions, ({ uuid: permission }) => {
+            if (permission.includes('*')) {
+                const [ application, resource, operation ] = permission.split(':');
+                return permissions.filter(p => (p.application === application
+                    && (resource === '*' || resource === p.resource)
+                    && (operation === '*' || operation === p.operation))
+                ).map(({ application, resource, operation }) => ({ uuid: `${application}:${resource}:${operation}` }));
+            }
+
+            return { uuid: permission };
+
+        })
+        : selectedPermissions);
 };
 
 const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...props }) => {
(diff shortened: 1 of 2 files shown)


src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -32,23 +32,42 @@ export const accessWrapper = (rawData, filters = { applications: [], resources:
...
         filteredData,
         applications: [ ...new Set(data.filter(permission => filterResource(permission) && filterOperation(permission)).map(({ application }) => application)) ],
         resources: [ ...new Set(data.filter(permission => filterApplication(permission) && filterOperation(permission)).map(({ resource }) => resource)) ],
         operations: [ ...new Set(data.filter(permission => filterApplication(permission) && filterResource(permission)).map(({ operation }) => operation)) ]
     };
 };
 
+export const resolveSplats = (selectedPermissions, permissions) => {
+    return (permissions.length > 0
+        ? selectedPermissions.map(({ uuid: permission }) => {
```
karelhala, line 45: I wouldn't use `map` and `flat` as some older browsers are not supporting it fully (hence the new depedency for core-js right?), I'd go with `flatMap` from lodash in this example `import flatMap from 'loadash/flatMap';`

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -32,23 +32,42 @@ export const accessWrapper = (rawData, filters = { applications: [], resources:
...
+};
+
 const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...props }) => {
     const dispatch = useDispatch();
     const fetchData = (apiProps) => dispatch(listPermissions(apiProps));
     const { access, isLoading, pagination, baseRole } = useSelector(selector, shallowEqual);
     const { input } = useFieldApi(props);
     const formOptions = useFormApi();
-    const [ permissions, setPermissions ] = useState({ filteredData: [], applications: [], resources: [], operations: []});
+    const [ permissions, setPermissions ] = useState({ data: [], filteredData: [], applications: [], resources: [], operations: []});
```
karelhala, line 66: Might be better to use `useReducer` in future, instead of `useState` with complex state, but that's just a nitpicking, might be good to mark it somehow as code cleenup and left as is.

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -80,13 +99,20 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...p
         setPermissions(accessWrapper(access, filters));
     }, [ filters ]);
 
+    useEffect(() => {
+        const resolvedPermissions = resolveSplats(selectedPermissions, permissions.data);
```
karelhala, line 103: No need to assign to a constant, also could be better to use state callback so we are in sync
```JS
setSelectedPermissions(() => resolveSplats(selectedPermissions, permissions.data));
```

karelhala: Looking at the second useEffect, it is better to combine effects together so we don't fire unwanted re-renders.
```JS
useEffect(() => {
  if (baseRole) {
    setSelectedPermissions(() => resolveSplats(
      baseRole.access?.map(({ permission  }) => ({ uuid: permission })),
      permissions.data
    ));
  } else {
    setSelectedPermissions(() => resolveSplats(selectedPermissions, permissions.data));
  }
}, [ permissions.data, baseRole ])
```

Here you can actually see that if there is any base role selected it will set selected permissions based on this and if not set it will always use the other function. With the previous implementation you would have to keep in mind that the selected permissions are being set from other effect and check which one takes precedens.

src/test/role/add-permission.test.js
```diff
@@ -1,4 +1,5 @@
-import { accessWrapper } from '../../smart-components/role/add-role-new/add-permissions';
+import 'core-js';
```
karelhala, line 1: With the `flatMap` you won't need `core-js` in here.

package-lock.json
```diff
@@ -5223,6 +5223,13 @@
       "integrity": "sha512-iT40nudw8zmCweivz6j58g+RT33I4KbaIvRUhjNmDwO2WmsQUxFEZZYZ5w3vXe5x5MX9D7mfvA/XaLOZYFR9EQ==",
       "requires": {
         "core-js": "^2.5.0"
+      },
+      "dependencies": {
```
karelhala, line 5227: Tento soubor asi ani nemusis pushovat.

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -32,23 +33,43 @@ export const accessWrapper = (rawData, filters = { applications: [], resources:
...
     const fetchData = (apiProps) => dispatch(listPermissions(apiProps));
     const { access, isLoading, pagination, baseRole } = useSelector(selector, shallowEqual);
     const { input } = useFieldApi(props);
     const formOptions = useFormApi();
-    const [ permissions, setPermissions ] = useState({ filteredData: [], applications: [], resources: [], operations: []});
+    // TODO: use reducer when cleaning this code
+    const [ permissions, setPermissions ] = useState({ data: [], filteredData: [], applications: [], resources: [], operations: []});
     const [ filters, setFilters ] = useState({ applications: [], resources: [], operations: []});
 
+ ;
```
karelhala, line 71: Extra `;`

karelhala (approved):

fhlavac (approved):

PR #430, opened 2020-08-31
Change chips into labels

PanSpagetka:
### Before
Screenshot from 2020-08-31 10-30-25
Screenshot from 2020-08-31 10-30-38

### After
Screenshot from 2020-08-31 11-14-21
Screenshot from 2020-08-31 11-14-32

### DIFF
--- a/src/smart-components/role/add-role-new/add-role-wizard.scss
+++ b/src/smart-components/role/add-role-new/add-role-wizard.scss
@@ -13,6 +13,11 @@
     display: block;
 }
 
-.ins-c-rbac-seleted-chips {
-    padding-bottom: var(--pf-global--spacer--md);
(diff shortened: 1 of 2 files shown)


mmenestr:
That looks great to me! Thanks for doing that so fast Robin

src/smart-components/role/add-role-new/add-role-wizard.scss
```diff
@@ -15,4 +15,7 @@
 
 .ins-c-rbac-seleted-chips {
     padding-bottom: var(--pf-global--spacer--md);
+    .pf-c-chip-group {
```
karelhala, line 18: Can you make the rule more specific so we don't have to write `!important`?

karelhala (approved):

PR #436, opened 2020-09-01
Enable add group steps after group name is valid

fhlavac:
https://projects.engineering.redhat.com/browse/RHCLOUD-8627

**BEFORE:**
"When creating a group, if a user navigates to the "Review" step immediately (without entering any group name) – they see the "Group name already taken" screen."

**NOW:**
Left nav is enabled only if group name is valid

@john-dupuy 
@mmenestr
@karelhala

### DIFF
--- a/src/smart-components/group/add-group/add-group-wizard.js
+++ b/src/smart-components/group/add-group/add-group-wizard.js
@@ -27,6 +27,7 @@ const AddGroupWizard = ({
   const [ formData, setValues ] = useState({});
   const [ isGroupInfoValid, setIsGroupInfoValid ] = useState(false);
   const [ hideFooter, setHideFooter ] = useState(false);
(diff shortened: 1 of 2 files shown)


john-dupuy:
Nice work @fhlavac! Testing this locally, I noticed that if a user is fast enough in clicking "Next" after entering a duplicate name (i.e. clicking it while the check is loading), that they could still navigate to the next section even though the left nav has been disabled.

I don't really think this will be an issue, as users won't hit this unless they're trying to break things. Just wanted to mention it though.

I'll attach a video to the ticket.

fhlavac:
@john-dupuy nice catch! fixed, thanks :-)

mmenestr:
I don't see a video, but i'll trust John and say it's good to go!

PanSpagetka (approved):

karelhala (approved):

PR #437, opened 2020-09-02
Fix disabled next on first step

PanSpagetka:
There is some validation bug described in  https://github.com/data-driven-forms/react-forms/issues/431 that is why `ValidationReset` is necessary.

https://projects.engineering.redhat.com/browse/RHCLOUD-8868


john-dupuy:
Checking this out and testing locally, the bug appears to be fixed, thanks @PanSpagetka!

karelhala (approved):

PR #442, opened 2020-09-03
Fix deselecting permissions on filter

PanSpagetka:
This issue was only present when creating a new custom role (not copying).

https://projects.engineering.redhat.com/browse/RHCLOUD-8870


john-dupuy (approved):
LGTM, I tested this and was able to see that the permissions were not deselected upon applying a filter.

karelhala (approved):

PR #444, opened 2020-09-07
Chrome filter fix

epwinchell:
This PR creates a Chrome-specific fix for the Users screen user name input

@karelhala I opened an issue here: https://github.com/patternfly/patternfly/issues/3449. This is also an issue in frontend-components, but I was not able to find another filter with placeholder text long enough to force cropping.  If we come across one, this is the issue.

Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-8941

Old
Screen Shot 2020-09-07 at 10 14 22 AM

New
Screen Shot 2020-09-07 at 10 13 58 AM

### DIFF
--- a/src/smart-components/user/user.scss
+++ b/src/smart-components/user/user.scss
@@ -1,3 +1,8 @@
 .ins-c-rbac__user-label-skeleton {
   width: 50px;
 }
+
+// fix for cropped placeholder: https://github.com/patternfly/patternfly/issues/3449
+ input#filter-by-username.ins-c-conditional-filter[type="text"] {
+  width: 230px;
+} 


epwinchell:
@karelhala Lint doesn't seem to like the Chrome selector

src/smart-components/user/user.scss
```diff
@@ -1,3 +1,9 @@
 .ins-c-rbac__user-label-skeleton {
   width: 50px;
 }
+
+// Chrome-specific fix for cropped  placeholder: https://github.com/patternfly/patternfly/issues/3449
+_:lang(x)::-internal-media-controls-overlay-cast-button, 
+input#filter-by-username.ins-c-conditional-filter[type="text"] {
```
karelhala, line 7: Good catch! I'd prefer here setting the `min-width` value for all browsers. It would look a bit off if user openes it in one browser and see the filter a bit smaller. Also any reason why to use `!important`? If we can avoid it it's always better to use more complex selectos

karelhala (approved):
Looking good!

src/smart-components/user/user.scss
epwinchell: @karelhala It worked without !important - done.

PR #450, opened 2020-09-09
Change use display_name insted of name for roles

PanSpagetka:
https://projects.engineering.redhat.com/browse/RHCLOUD-8089


karelhala (commented):
What if `display_name` is empty but we see `name`? I think it would be nice to users to show either `display_name || name` in these cases so they don't see empty cells.

PR #455, opened 2020-09-10
Replace "ins-c-rbac__dialog--warning" class with React token

epwinchell:
Old
Screen Shot 2020-09-10 at 2 41 29 PM

New
Screen Shot 2020-09-10 at 2 35 23 PM

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -38,12 +38,6 @@
   }
 }
 
-.ins-c-rbac__dialog--warning {
-  .ins-c-rbac__delete-icon {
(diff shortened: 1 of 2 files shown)


karelhala:
AFAIK this is the same, but imported from JS, right? We prefer CSS variables over JS one's as they can be rewritten, but if there are some benefits let me know.

epwinchell:
@karelhala It's the same, but as a modal header - not an example of something that would likely need to be rewritten. I'm fine with it either way.

PR #459, opened 2020-09-15
Make user access bundle cards clickable.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-8857

### TODO: 
- [x] more test
- [x] add some pictures

### Org admin view
mua-org-admin

### User view
mua-user

cc @beccaglass


epwinchell (commented):
Looks good. I did notice that when you first hit the page, the active card as a border on top, but after clicking on a card, the active card has a border all the way around.

Initial active state
Screen Shot 2020-09-18 at 11 35 26 AM


After selecting
Screen Shot 2020-09-18 at 11 35 43 AM

beccaglass:
@epwinchell good catch! Ideally would just have the top border whenever selected

Hyperkid123:
@epwinchell @beccaglass that is a default browser outline highlight on any focusable, in this case clickable, element existing for accessibility reasons. If you want that to change I would suggest contacting the PF design team to come up with some pattern for the CSS outline property.

We can theoretically turn it off by setting `outline: none` but that will basically destroy any accessibility help on that particular element.

Hyperkid123:
Here is a short GIF to show that I am not making it up :smile: 

PF-outline

epwinchell:
@Hyperkid123 We never doubted you 😄

epwinchell (approved):

PR #462, opened 2020-09-16
Make default access group sticky

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-8202
https://marvelapp.com/prototype/129585j4

- made default access group sticky
- added popover with default group description
- updated tests

@mmenestr


mmenestr:
Looks good! Just one small thing, could you make the popover wider so that it just becomes two lines, and default isn't alone on the third line haha
Also, is there an issue in the works, to change the name of the group from "Custom default user access" to just "custom default access"?

ryelo:
Yeah, the issue to rename the group should be on https://issues.redhat.com/browse/RHCLOUD-8201 which we are blocked on the API on right now.

fhlavac:
Ok, made the popover wider. It should work for both versions of default group name.

mmenestr:
looks good to go then!

src/redux/actions/group-actions.js
```diff
@@ -6,6 +6,14 @@ export const fetchGroups = (options = {}) => ({
   payload: GroupHelper.fetchGroups(options)
 });
 
+export const fetchSystemGroup = (filterValue) => ({
+  type: ActionTypes.FETCH_SYSTEM_GROUP,
+  payload: GroupHelper.fetchGroups({
+    name: filterValue !== '' ? filterValue : 'default',
```
karelhala, line 12: If the `filterValue` is undefined it will be used in here. Is that intentional? I think it would be better to use shortcutting.
```JS
name: filterValue || 'default'
```

Or even better
```JS
export const fetchSystemGroup = (filterValue = 'defaul') => ({
```

src/redux/reducers/group-reducer.js
```diff
@@ -24,6 +25,7 @@ const setLoadingState = state => ({ ...state, isLoading: true });
 const setRecordLoadingState = state => ({ ...state, isRecordLoading: true, selectedGroup: { ...state.selectedGroup, loaded: false }});
 const setRecordRolesLoadingState = state => ({ ...state, isRecordRolesLoading: true, selectedGroup: { ...state.selectedGroup, loaded: false }});
 const setGroups = (state, { payload }) => ({ ...state, groups: payload, isLoading: false });
+const setSystemGroup = (state, { payload }) => ({ ...state, systemGroup: payload.data.filter(group => group.platform_default)[0] });
```
karelhala, line 28: Conditional accessor might be good here so we don't break the UI
```suggestion
const setSystemGroup = (state, { payload }) => ({ ...state, systemGroup: payload?.data?.filter(group => group?.platform_default)?.[0] });
```

src/smart-components/group/group-table-helpers.js
```diff
@@ -1,27 +1,47 @@
...
+            </Link>
+            { isPlatformDefault &&
+            <Popover
+              zIndex="110"
+              position="right"
+              className="pf-c-default-group-popover"
+              isVisible={ isPopoverVisible }
+              shouldClose={ () => setPopoverVisible(false) }
+              hideOnOutsideClick={ true }
+              bodyContent={ 'This group contains the roles that all users in your organization inherit by default.' }
```
karelhala, line 29: No need to wrap it in object.
```suggestion
              bodyContent="This group contains the roles that all users in your organization inherit by default."
```

src/smart-components/group/groups.js
```diff
@@ -30,16 +30,23 @@ const Groups = () => {
   const [ removeGroupsList, setRemoveGroupsList ] = useState([]);
 
   const dispatch = useDispatch();
-  const { groups, pagination, userIdentity, isLoading } = useSelector(({ groupReducer: { groups, isLoading }}) => ({
-    groups: groups.data,
+  const { groups, pagination, userIdentity, isLoading } = useSelector(({ groupReducer: { groups, isLoading, systemGroup }}) => ({
+    groups: groups.data && systemGroup ?
```
karelhala, line 34: I don't think you need to perform this check
```JS
groups: [
  ...(systemGroup?.name?.toLowerCase()?.includes(filterValue) ? [ systemGroup ] : []),
  ...groups?.data?.filter(({ platform_default } = {}) => !platform_default) || []
]
```

karelhala: @mmenestr I don't remeber what we agreed on, do we want to show or hide default group when filtering? This will hide it, which is totally fine if we want to do that. Just wanted to be sure about it.

src/smart-components/group/groups.js
```diff
@@ -30,16 +30,23 @@ const Groups = () => {
...
     userIdentity: groups.identity,
-    isLoading
+    isLoading,
+    systemGroup
   }), shallowEqual);
 
   useEffect(() => {
     insights.chrome.appNavClick({ id: 'groups', secondaryNav: true });
     dispatch(fetchGroups({ ...pagination, name: filterValue }));
+    dispatch(fetchSystemGroup(name));
```
karelhala, line 49: Missing `filterValue`
```suggestion
    dispatch(fetchSystemGroup(filterValue));
```

karelhala (commented):
Looking good! Just a few minor improvements and we are golden.

src/smart-components/group/groups.js
fhlavac: @karelhala @mmenestr current implementations shows the default group on filter only if it matches given filterValue. That's what I thought to be expected.

fhlavac:
I reset my branch by mistake - moved everything to the new PR
https://github.com/RedHatInsights/insights-rbac-ui/pull/469

PR #466, opened 2020-09-17
Add prettier

Hyperkid123:
The code formating was all over the place in this repo. I have talked to @karelhala and I want to introduce prettier to keep an eye on us.

Plus the prettier plugin has formatters for all its rules so we can just use the eslint --fix option to do all the work for us :astonished:


ryelo:
Sweet! Is this something we'd want to roll into the starter app too?

Hyperkid123:
@ryelo It does make the code prettier IMO. I am totally for it.

epwinchell (approved):

PR #469, opened 2020-09-18
Make default access group sticky

fhlavac:
New version of https://github.com/RedHatInsights/insights-rbac-ui/pull/462

https://issues.redhat.com/browse/RHCLOUD-8202
https://marvelapp.com/prototype/129585j4

- made default access group sticky
- added popover with default group description
- updated tests

### DIFF
--- a/src/smart-components/group/groups.scss
+++ b/src/smart-components/group/groups.scss
@@ -1,3 +1,20 @@
 .pf-c-table tbody tr.ins-c-rbac__group-default {
   .pf-c-table__check:first-of-type input { display: none; }
 }
+
+.pf-c-question-circle-icon {
+  color: var(--pf-global--palette--black-600);
+  margin-left: var(--pf-global--spacer--sm);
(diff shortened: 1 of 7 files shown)


src/smart-components/group/groups.scss
```diff
@@ -1,3 +1,20 @@
...
+    color: var(--pf-global--palette--blue-400)
+  }
+}
+
+.pf-c-popover {
+  min-width: 350px;
+}
+
+body {
+  overflow: hidden
```
Hyperkid123, line 19: I don't like this rule. It might mess up the whole layout. Can you add the overflow to a different element? Is this the "fix" for messy tables with popovers?

fhlavac: fixed with appendTo property, thanks for help @Hyperkid123

Hyperkid123 (approved):

karelhala (approved):

PR #476, opened 2020-09-21
Make default group filtering case insensitive

fhlavac:
BEFORE:
"Search is case-sensitive. If I search for "custom", the group shows up, if I search for "Custom", it does not."

NOW:
Default access group shows up in both cases

@john-dupuy


john-dupuy:
Tested locally and this fixed the issue! Thanks @fhlavac :+1:

john-dupuy (approved):

karelhala (approved):

PR #478, opened 2020-09-22
Custom roles edit

Hyperkid123:
tasks:
- https://issues.redhat.com/browse/RHCLOUD-8948
- https://issues.redhat.com/browse/RHCLOUD-8947

update-remove-role

### TODO
- [x] images
- [x] tests
- [x] ~~allow actions only for org admins~~ (will do this later once RBAC has opened to non-org admins)


src/helpers/role/role-helper.js
```diff
@@ -38,3 +38,5 @@ export async function fetchRoleForPrincipal(uuid) {
 export async function removeRole(roleId) {
   return await roleApi.deleteRole(roleId);
 }
+
+export const updateRole = (roleId, data) => roleApi.getRoleAccess(roleId).then(({ data: access }) => roleApi.updateRole(roleId, { ...data, access }));
```
karelhala, line 42: Can you use `async/await` in here?
```suggestion
export const updateRole = async (roleId, data) => {
  const { data: access } = await roleApi.getRoleAccess(roleId);
  return roleApi.updateRole(roleId, { ...data, access });
};
```

Hyperkid123: @karelhala  :rage: :smile:  OK.

Just to clarify what is going on here. I do plan to open a ticket on the API because this can potentially cause issues. Access is required inside the put requests but if we query to access data like it will return 100 items max. And you can intentionally truncate the access data when updating. I think the update endpoint should not have access attributes required.

karelhala (approved):

PR #480, opened 2020-09-22
Grouping error

arivepr:
Fixed bug where the edit modal would not show up for the selected group(s).


john-dupuy:
Verified that this fix works, thanks @arivepr!

john-dupuy (approved):

src/smart-components/group/groups.js
```diff
@@ -70,7 +70,7 @@ const Groups = () => {
           }}
         />
       </Route>
-      <Route exact path={paths['group-edit']}>
+      <Route exact path={ paths['group-edit'].path }>
```
karelhala, line 73: Remove the extra spaces so lint is fixed and we are good to go!
```suggestion
      <Route exact path={paths['group-edit'].path}>
```

karelhala (approved):
Looking good!

ryelo (approved):

PR #490, opened 2020-09-24
MUA mobile layout

epwinchell:
jira: https://issues.redhat.com/browse/RHCLOUD-9193

This PR uses the responsive grid to handle both the desktop and new mobile layout, per the mocks here: https://marvelapp.com/prototype/11gc312j/screen/72960978

Follow-up items:
- the dropdown in the header is a placeholder 
- the bottom toolbar (pagination) is broken in RBAC
- Table Toolbar filter should work just like this example: https://www.patternfly.org/v4/documentation/react/demos/filtertabledemo#component-title with  filter icon as a toggle in the mobile layout

"Sticky" Elements

I ran into a challenging situation: sticky elements can't be nested.  They have to stick to their parent - any hierarchy has to be squashed:

Original :

```
<Section>
  <Header>		                    
     <Title>
         My User Access
        <span>Understand your Red Hat access...</span>
     </Title>
  </Header>
</Section>
<Section>
  Rest of the screen
</Section>
```

This PR:
```
<sticky title />
<non-sticky subtitle />
<sticky dropdown />

<Section>
  Rest of the screen
</section>
```

Screen Shot 2020-09-27 at 9 37 25 AM

@karelhala @Hyperkid123  Since we are likely to be creating complex mobile layouts going forward, how about we investigate using [react-sizeme](https://github.com/ctrlplusb/react-sizeme)  or [react-measure](https://github.com/souporserious/react-measure), so we can render different elements or completely different layouts depending on screen size rather than hacking to make something like sticky elements work?

### DIFF
--- a/src/smart-components/myUserAccess/MUAContent.js
+++ b/src/smart-components/myUserAccess/MUAContent.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Stack, StackItem, Title } from '@patternfly/react-core';
+import { Grid, GridItem, Stack, StackItem, Title } from '@patternfly/react-core';
 
 import MUACard from '../../presentational-components/myUserAccess/MUACard';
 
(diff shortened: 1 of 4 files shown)


Hyperkid123 (approved):

Hyperkid123:
> @karelhala @Hyperkid123 Since we are likely to be creating complex mobile layouts going forward, how about we investigate using react-sizeme or react-measure, so we can render different elements or completely different layouts depending on screen size rather than hacking to make something like sticky elements work?

Looking briefly at the source code, the react-sizeme looks like total overkill and I would not be surprised if it would come with a noticeable performance hit mainly on smaller devices. The other option looks better IMO.

On the other hand, if there will be major layout differences between the two versions, we can have different lazy loaded "entry points" and have just switch them when we hit the breakpoint. Realistically I don't think that there are users that are switching between layouts regularly and "splitting the app in two versions" and using a reusable responsive component might be easier than having multiple checks (with or without library) on each page and basically duplicating and loading each layout twice. We are not talking about duplicating any logic or anything so it probably would not take that much time. And we would be able to achieve this with one simple private observer instead of having many and assigning then to global window objects like these libraries are doing.

Hyperkid123 (approved):

PR #492, opened 2020-09-25
Set max width to add permissions filter

PanSpagetka:
Filter select was too wide and pushed pagination on next row.

Before
Screenshot from 2020-09-25 10-08-24

After
Screenshot from 2020-09-25 10-07-52

@mmenestr


src/smart-components/role/add-role-new/add-role-wizard.scss
```diff
@@ -40,4 +40,20 @@
...
 
-
+.ins-c-rbac-permissions-table {
+    .ins-c-primary-toolbar {
+        .pf-c-toolbar__content {
+            .pf-c-toolbar__content-section {
+                .pf-m-filter-group {
+                    .ins-c-primary-toolbar__filter {
+                        .ins-c-conditional-filter {
+                            .pf-c-select {
```
karelhala, line 50: Phew, this is really complex. Isn't it possible to simplify it a bit?

PanSpagetka: I didn't want to change the Toolbar component, so there is quite complex selection, but I think I can simplify it a bit.

mmenestr:
Perfect, thank you!!

src/smart-components/role/add-role-new/add-role-wizard.scss
```diff
@@ -42,3 +42,15 @@
         }
     }
 }
+
+.ins-c-rbac-permissions-table {
+    .ins-c-primary-toolbar {
+        .pf-c-toolbar__content {
+            .pf-c-toolbar__content-section {
+                .pf-m-filter-group {
+                    max-width: 35em;
```
Hyperkid123, line 51: How responsive is this? I think that using a % value might be a bit safer.

PanSpagetka: I have experiment a bit with media queries and it seems to be working well now. Even though it is a bit complex. 
Unfortunately, github does not support `.webm` files https://webm.red/view/YtI4.webm
(I can't make gifs)

Hyperkid123: Yeah that looks good. Can you fix up travis?

Hyperkid123 (approved):

PR #498, opened 2020-09-29
Use api to get filter options for Add permissions

PanSpagetka:


src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -94,9 +113,46 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...p
       dispatch(fetchRole(baseRoleUuid));
     }
 
+    debounbcedGetApplicationOptions = debounce(
+      ({ applications, resources, operations }) =>
+        fetchOptions({ field: 'application', limit: 50, application: applications.join(), resourceType: resources.join(), verb: operations.join() }),
```
karelhala, line 118: You should use `debouncePromise ` from `@redhat-cloud-services/frontend-components-utils` instead of `debounce`

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -94,9 +113,46 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...p
       dispatch(fetchRole(baseRoleUuid));
     }
 
+    debounbcedGetApplicationOptions = debounce(
```
karelhala, line 116: You should use `useCallback` instead of assigning it in `useEffect` it should be slightly faster and more readable.

package.json
```diff
@@ -15,7 +15,7 @@
     "@redhat-cloud-services/frontend-components": "2.4.5",
     "@redhat-cloud-services/frontend-components-notifications": "2.1.2",
     "@redhat-cloud-services/frontend-components-utilities": "2.1.1",
-    "@redhat-cloud-services/rbac-client": "^1.0.73",
+    "@redhat-cloud-services/rbac-client": "^1.0.74",
```
karelhala, line 18: You might have forgotten to include `package-lock.json`

Hyperkid123 (approved):

Hyperkid123:
Just a not. I think this component requires some refactoring. These 100 effects in one component might get us in trouble.

PR #510, opened 2020-10-06
Add ouiaId's to several components

john-dupuy:
Adding ouiaId's to several components in the RBAC frontend. Some will have to be added in the `frontend-components` repo. 

Or we need some way to pass the ouiaId props up to those components.

### DIFF
--- a/src/smart-components/role/role-permissions.js
+++ b/src/smart-components/role/role-permissions.js
@@ -100,7 +100,7 @@ const Permissions = () => {
           isDisabled: !selectedPermissions.length > 0,
         },
         onClick: () => {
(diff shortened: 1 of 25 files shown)


src/smart-components/role/role-permissions.js
```diff
@@ -100,7 +100,7 @@ const Permissions = () => {
           isDisabled: !selectedPermissions.length > 0,
         },
         onClick: () => {
-          const multiplePermissionsSelected = selectedPermissions.lenght > 1;
+          const multiplePermissionsSelected = selectedPermissions.length > 1;
```
karelhala, line 103: Good catch!

karelhala (approved):
Looking good!

PR #512, opened 2020-10-07
Add resource definiton step to Custom Role wizard

PanSpagetka:
Mapping between `resource type` part of permission to url and back is a bit janky.
  
https://issues.redhat.com/browse/RHCLOUD-8349

Screenshot from 2020-10-13 14-27-02
Screenshot from 2020-10-13 14-27-14


src/helpers/shared/user-login.js
```diff
@@ -36,3 +40,10 @@ export function getPermissionApi() {
 export function getAxiosInstance() {
   return axiosInstance;
 }
+
+export function getCostApi() {
+  return {
+    getResourceTypes: () => costApi.axios.get(`${COST_API_BASE}/resource-types/`),
```
Hyperkid123, line 46: I guess the cost API does not expose these endpoints?

src/smart-components/role/add-role-new/cost-resources.js
```diff
@@ -0,0 +1,171 @@
...
+      }),
+      {}
+    )
+  );
+  const onToggle = (key, isOpen) => dispatchLocaly({ type: 'toggle', key, isOpen });
+  const clearSelection = (key) => dispatchLocaly({ type: 'clear', key });
+  const onSelect = (event, selection, selectAll, key) =>
+    selectAll ? dispatchLocaly({ type: 'selectAll', selection, key }) : dispatchLocaly({ type: 'select', selection, key });
+
+  const permissionToResource = (permission) => resourceTypes.find((r) => r.value === permission.split(':')[1])?.path.split('/')[5];
```
Hyperkid123, line 92: Could you add a conditional chain to the array entries as well? `foo?.[0]`

src/smart-components/role/add-role-new/cost-resources.js
```diff
@@ -0,0 +1,171 @@
...
+    formOptions.change('resource-definitions', resourceDefinitions);
+  }, [state]);
+
+  // eslint-disable-next-line react/prop-types
+  const makeRow = ({ uuid: permission }) => {
+    const options = resources[permissionToResource(permission)] || [];
+    return (
+      <React.Fragment>
+        <GridItem span={3}>
+          <FormGroup label={permission} isRequired></FormGroup>
```
Hyperkid123, line 125: How does this behave on small screens?
```jsx
<FormGroup label={permission} isRequired />
```

src/helpers/shared/user-login.js
PanSpagetka: Kind of, their schema is broken and and they are not using JS client.

src/smart-components/role/add-role-new/cost-resources.js
PanSpagetka: I would say that it is ok
Screenshot from 2020-10-13 10-05-40

Hyperkid123: I think it would be worthwhile to switch to a horizontal layout on small screens. It would definitely improve the mobile experience. The labels are just way too long.

PanSpagetka: Yep, it is looks better. I have added screenshots to the description.

epwinchell: @PanSpagetka @Hyperkid123  I think we need input from @mmenestr on what the mobile layout for this should look like. (I can't see repeating the labels on every line.) Anyway, this modal doesn't work in mobile at the moment anyway because the steps should switch to a dropdown in the header (Jira issue: https://issues.redhat.com/browse/RHCLOUD-9548)

 I say merge this and we'll deal with the mobile issues later.

Hyperkid123 (approved):

src/smart-components/role/add-role-new/cost-resources.js
epwinchell: Screen Shot 2020-10-13 at 9 55 07 AM

PR #521, opened 2020-10-08
Edit resource definitions

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-8353

- added displaying the "Resource definitions" column for cost roles
- added resource definitions detail
- added resource definition edit modal
- enhanced tests

@mmenestr


src/smart-components/common/ModalFormTemplate.js
```diff
@@ -53,8 +53,9 @@ CustomButtons.defaultProps = {
   cancelLabel: 'Cancel',
 };
 
-const ModalFormTemplate = ({ ModalProps, saveLabel, cancelLabel, ...props }) => (
+const ModalFormTemplate = ({ ModalProps, saveLabel, cancelLabel, alert, ...props }) => (
   <Modal {...ModalProps} footer={<CustomButtons saveLabel={saveLabel} cancelLabel={cancelLabel} />}>
+    {alert && alert()}
```
karelhala, line 58: Can use safe accesor.
```suggestion
    {alert?.()}
```

src/smart-components/role/edit-resource-definitions-modal.js
```diff
@@ -0,0 +1,193 @@
...
+  const [loadingStateVisible, setLoadingStateVisible] = useState(true);
+  const [changedResources, setChangedResources] = useState();
+
+  const { definedResources, role } = useSelector(
+    (state) => ({
+      role: state.roleReducer.selectedRole,
+      definedResources: state.roleReducer.selectedRole?.access
+        ? state.roleReducer.selectedRole.access
+            .find((a) => a.permission === permissionId)
+            .resourceDefinitions.reduce((acc, curr) => [...acc, curr.attributeFilter.value], []) // tady to jsem přehodil komentář, nefunguje!
```
karelhala, line 79: Uuum, can we remove this? :smile:

src/smart-components/role/edit-resource-definitions-modal.js
```diff
@@ -0,0 +1,193 @@
...
+        push(cancelRoute);
+      }
+    );
+  };
+
+  const options = createOptions(resources);
+
+  return (
+    <React.Fragment>
+      <WarningModal
```
karelhala, line 140: No need to wrap in `React.Fragment`

src/smart-components/role/role-permissions-table-helpers.js
```diff
@@ -13,6 +14,17 @@ export const createRows = (data, opened, selectedRows = []) =>
           appName,
           type,
           operation,
+          ...(showResDefinitions
+            ? [
+                permission.includes('cost-management') && resourceDefinitions.length > 0 ? ( // change to > when final!
```
karelhala, line 19: Is the comment neccessary?

karelhala:
This is relatively big PR, I've gone trough the code anc couldn't find anything obious, might want to test this locally as well.

src/smart-components/common/warningModal.js
```diff
@@ -4,12 +4,12 @@ import { Modal, ModalVariant, Button } from '@patternfly/react-core';
...
 import './warningModal.scss';
 
-export const WarningModal = ({ type, isOpen, onModalCancel, onConfirmCancel }) => (
+export const WarningModal = ({ type, isOpen, onModalCancel, onConfirmCancel, customTitle, customDescription }) => (
   <Modal
     title={
       <span className="ins-c-wizard__cancel-warning-header">
         <ExclamationTriangleIcon size="md" className="ins-c-wizard__cancel-warning-header--icon" />
-        Exit {type} creation?
+        {customTitle ? customTitle : `Exit ${type} creation?`}
```
rvsia, line 12: You can `customTitle || ...`

src/smart-components/common/warningModal.js
```diff
@@ -25,12 +25,14 @@ export const WarningModal = ({ type, isOpen, onModalCancel, onConfirmCancel }) =
       </Button>,
     ]}
   >
-    <span> All inputs will be discarded.</span>
+    <span>{customDescription ? customDescription : 'All inputs will be discarded.'} </span>
```
rvsia, line 28: ^^

src/smart-components/common/warningModal.js
```diff
@@ -25,12 +25,14 @@ export const WarningModal = ({ type, isOpen, onModalCancel, onConfirmCancel }) =
...
   </Modal>
 );
 
 WarningModal.propTypes = {
   type: PropTypes.string,
+  customTitle: PropTypes.string,
```
rvsia, line 34: Can be `node` if there is no specific reason why it is a string

src/smart-components/role/edit-resource-definitions-modal.js
```diff
@@ -0,0 +1,193 @@
...
+    }
+  }, [resourceTypes]);
+
+  const onCancel = () => replace(cancelRoute);
+
+  const handleCancel = (data) => {
+    if (data['dual-list-select'] === definedResources) {
+      onCancel();
+    } else {
+      setChangedResources(data['dual-list-select']);
```
rvsia, line 111: Suggestion: multiple useState calls can be wrapped to one useReducer, it will prevent additional renders

src/smart-components/role/role-permissions.js
```diff
@@ -22,7 +22,28 @@ import { removeRolePermissions, fetchRole } from '../../redux/actions/role-actio
 
 const maxFilterItems = 10;
 
-const columns = [{ title: 'Application' }, { title: 'Resource type' }, { title: 'Operation' }, { title: 'Last commit', transforms: [cellWidth(15)] }];
+const resourceDefinitionsTitle = 'Resource definitions';
```
rvsia, line 25: I think this const is unneccesary, can be set right in the object

src/smart-components/role/edit-resource-definitions-modal.js
fhlavac: I think it's needed since I have 2 elements - WarningModal and Modal/FormRenderer below

mmenestr:
I just have a few little things:
1.  Could you change the "Resources added to the permission" heading in the second box, to "Resources **defined for** the permission"?
2. Could you change the "submit" button to "save"?
3. Could you make the search bar wider so that the search + sort is the same width as the card width?

Otherwise looks great!

src/smart-components/role/edit-resource-definitions-modal.js
```diff
@@ -0,0 +1,193 @@
...
+import { getResource, getResourceDefinitions } from '../../redux/actions/cost-management-actions';
+import componentMapper from '@data-driven-forms/pf4-component-mapper/dist/cjs/component-mapper';
+import { WarningModal } from '../common/warningModal';
+import { Spinner, Modal } from '@patternfly/react-core';
+import ResourceDefinitionsFormTemplate from './ResourceDefinitionsFormTemplate';
+import './role-permissions.scss';
+
+const createOptions = (resources) => {
+  let options = [];
+  for (const [key, value] of Object.entries(resources)) {
```
PanSpagetka, line 18: I think we should avoid using for loops unless there is specific reason to use it.
```
Object.entries(resources).reduce((acc, [key, value]) => [...acc, value.map((r) => ({ value: r.value, path: key, label: r.value }))], []);
```

src/smart-components/role/role-permissions.scss
```diff
@@ -5,4 +5,21 @@
...
     }
   }
+
+  .ins-c-text__disabled {
+    color: var(--pf-global--Color--200);
+  }
+}
+
+.ins-m-resource-definitions {
+  .pf-c-modal-box__body {
```
PanSpagetka, line 15: If you can find PF variables with same values that would be cool :) If not that is also fine

src/smart-components/role/role-resource-definitions.js
```diff
@@ -0,0 +1,148 @@
...
+  const [config, setConfig] = useState({
+    pagination: {
+      ...defaultSettings,
+      filter: '',
+    },
+  });
+
+  const { pagination, filter } = config;
+
+  const columns = [{ title: 'Title' }];
```
PanSpagetka, line 26: Shouldn't be here some more specific string?

PanSpagetka (changes requested):

src/smart-components/role/edit-resource-definitions-modal.js
```diff
@@ -0,0 +1,193 @@
...
+      filterValueTitle: 'Filter by resource...',
+      options: [...(resourcesPath && resources ? options : [])],
+      validate: [{ type: 'validate-resources' }],
+    },
+  ],
+});
+
+const selector = ({ costReducer: { resourceTypes, isLoading, loadingResources, resources } }, resourcesPath) => ({
+  resourceTypes: resourceTypes.data,
+  resources: resources[resourcesPath] ? { resourcesPath: resources[resourcesPath] } : resources,
```
PanSpagetka, line 49: I am pretty sure, that this piece of code is not doing what you want. You are creating object with `"resourcePath"` string as a key, that is not necessary. You can just select resources for current permission a be good with it. Also you should change `createOptions` function.
```
resources: resources[resourcesPath] || [],
```

src/smart-components/role/edit-resource-definitions-modal.js
PanSpagetka: Actually when you make changes to the selector, this can be:
```
const createOptions = (resources) => resources.map((r) => ({ value: r.value, label: r.value }));
```

src/smart-components/role/role-resource-definitions.js
fhlavac: > Shouldn't be here some more specific string?

I don't think it matters, the table header isn't visible. It just doesn't work for me when the name wasn't defined at all. Actually I can put empty object instead of this to columns

src/smart-components/role/edit-resource-definitions-modal.js
fhlavac: @rvsia Thanks for the advice! Just please take one more look if I am using it properly, thank you :)

src/smart-components/role/role-permissions.scss
fhlavac: replaced existing ones, thanks @PanSpagetka

PanSpagetka (approved):
Looks good to me

fhlavac:
> I just have a few little things:
> 
> 1. Could you change the "Resources added to the permission" heading in the second box, to "Resources **defined for** the permission"?
> 2. Could you change the "submit" button to "save"?
> 3. Could you make the search bar wider so that the search + sort is the same width as the card width?
> 
> Otherwise looks great!

No problem, I used these labels since they were in the mockups. It's fixed, thank you :)

PanSpagetka (approved):

PR #525, opened 2020-10-09
Use api for filtering

PanSpagetka:
Simplify code a bit and fix bugs.

https://issues.redhat.com/browse/RHCLOUD-8869


package.json
```diff
@@ -15,7 +15,8 @@
     "@redhat-cloud-services/frontend-components": "2.4.9",
     "@redhat-cloud-services/frontend-components-notifications": "2.1.2",
     "@redhat-cloud-services/frontend-components-utilities": "2.1.1",
-    "@redhat-cloud-services/rbac-client": "^1.0.77",
+    "@redhat-cloud-services/rbac-client": "^1.0.78",
+    "axios-mock-adapter": "^1.18.1",
```
Hyperkid123, line 19: Can you remove this dependency? We will not be using tests anymore. Instead, we will be mocking async requests via jest mock API.

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -20,45 +20,19 @@ const selector = ({
   },
   roleReducer: { isRecordLoading, selectedRole },
 }) => ({
-  access: permission.data,
+  permissions: permission.data.map((p) => ({
+    application: p.application,
+    resource: p.resource_type,
+    operation: p.verb,
+    uuid: p.permission,
+  })),
```
karelhala, line 28: Can you style it a bit better?
```suggestion
  permissions: permission.data.map(({ application, resource_type: resource, verb, permission } = {}) => ({
    application,
    resource,
    operation: verb,
    uuid: permission,
  })),
```

karelhala (commented):
Just one minor change and we are golden!

Hyperkid123 (approved):

PR #530, opened 2020-10-13
Fix single filter delete

PanSpagetka:
At Add Permissions step in Custom roles wizard, when you add filter and then try to remove it by clicking in X button in the chip. Filter is not removed.

https://issues.redhat.com/browse/RHCLOUD-9944


epwinchell (approved):
Verified fix.

PR #542, opened 2020-10-19
Mua redirect

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-9881

The users route was eating away the root route.

@ryelo @karelhala is there any additional rule for the MUA route? Is it supposed to be the root only on beta or in every environment? We can obviously change the root route based on that.

Also the lock file change should happen because there is no longer an axios mock adapter library. I guess it was omitted in some previous PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5343,27 +5343,6 @@
         "follow-redirects": "1.5.10"
       }
     },
-    "axios-mock-adapter": {
-      "version": "1.18.2",
-      "resolved": "https://registry.npmjs.org/axios-mock-adapter/-/axios-mock-adapter-1.18.2.tgz",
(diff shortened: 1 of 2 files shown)


karelhala (approved):

ryelo:
As of right now here are the routes we're concerned about:
`ci+qa/beta/settings` -> redirect to MUA
`ci+qa/beta/settings/my-user-access` -> MUA
`ci+qa/beta/settings/rbac` -> redirect to Rbac Users

These routes will be sent to prod-beta on Oct 29

Hyperkid123:
@ryelo alright. I will add isBeta check. Because now it just defaults to MUA everywhere.

PR #545, opened 2020-10-20
Add Submit to Create custom role

PanSpagetka:
Add submit to Create custom role.

Issue 6 from https://issues.redhat.com/browse/RHCLOUD-9983

### DIFF
--- a/src/smart-components/role/add-role-new/add-role-wizard.js
+++ b/src/smart-components/role/add-role-new/add-role-wizard.js
@@ -6,7 +6,7 @@ import FormRenderer from '@data-driven-forms/react-form-renderer/dist/cjs/form-r
 import Pf4FormTemplate from '@data-driven-forms/pf4-component-mapper/dist/cjs/form-template';
 import componentMapper from '@data-driven-forms/pf4-component-mapper/dist/cjs/component-mapper';
 import schemaBuilder from './schema';
-import { createRole, fetchRoles } from '../../../redux/actions/role-actions';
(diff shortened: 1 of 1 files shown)


src/smart-components/role/add-role-new/add-role-wizard.js
```diff
@@ -57,6 +57,40 @@ const AddRoleWizard = ({ history: { push } }) => {
...
+                  value: resourceDefinitions?.find((r) => r.permission === permission).resources,
+                },
+              },
+            ]
+          : [],
+      })),
+    };
+    const role = await dispatch(createRole(roleData));
+    fetchRoles().then(push('/roles'));
+    return role;
```
rvsia, line 91: Haven't tested it in UI, but it this seems like that after a click on the submit, the form is still opened until roles are fetched, so it would be possible to click multiple times.. is that right?

karelhala: This feels a bit overcomplicated, it will create new role, waits untill that is done, starts fetching roles, fills in the role to form and once the roles are fetched it redirects to said page. We should probably wait for the role to create and if it fails show user what happens, so we'll wait for the role once that succeeds start fetching roles, don't wait for it to happen and redirect to roles. It will redirect and once that is fetched it'll show the full table.

PanSpagetka: I think, that your concerns should be resolved in second commit.

rvsia (approved):

PR #548, opened 2020-10-22
Use create group API error response.

Hyperkid123:
jria: https://issues.redhat.com/browse/RHCLOUD-8736

### Changes
- remove ambiguous create group rejected notification
- format create group error API response to match notifications requirements
- format create group API error for unique name

### DIFF
--- a/src/redux/actions/group-actions.js
+++ b/src/redux/actions/group-actions.js
@@ -22,25 +22,23 @@ export const fetchGroup = (apiProps) => ({
 
 export const addGroup = (groupData) => ({
   type: ActionTypes.ADD_GROUP,
-  payload: GroupHelper.addGroup(groupData),
-  meta: {
-    notifications: {
-      fulfilled: {
-        variant: 'success',
-        title: 'Success adding group',
-        dismissDelay: 8000,
-        dismissable: false,
-        description: 'The group was added successfully.',
-      },
-      rejected: {
-        variant: 'danger',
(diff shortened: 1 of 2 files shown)


src/redux/actions/group-actions.js
```diff
@@ -22,7 +22,24 @@ export const fetchGroup = (apiProps) => ({
 
 export const addGroup = (groupData) => ({
   type: ActionTypes.ADD_GROUP,
-  payload: GroupHelper.addGroup(groupData),
+  payload: GroupHelper.addGroup(groupData).catch((err) => {
```
karelhala, line 25: Couldn't this be axios plugin?

Hyperkid123: Its gonna have to change. This event should replace the shole wizard step. The notification won't suffice. I got my hands on this: https://marvelapp.com/prototype/237ah31g/screen/71426702

karelhala: Mhh it could be an plugin, but meh.

Hyperkid123: So it probably won't be possible. This is a very specific scenario. Sometimes we need to prevent the error notification and consume the outcome in the wizard. You can see the code. Now we can add an interceptor that would create a message format that will not kill the UI, but that is probably for a different PR.

Hyperkid123:
rbac-name

john-dupuy (approved):

PR #555, opened 2020-10-26
Call chrome app action on root routes.

Hyperkid123:


src/smart-components/common/page-action-route.js
```diff
@@ -0,0 +1,16 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Route } from 'react-router-dom';
+
+const PageActionRoute = ({ pageAction, ...props }) => {
+  useEffect(() => {
+    insights.chrome.appAction(pageAction);
```
karelhala, line 7: You should clean once the route unmounts as well `return () => insights.chrome.appAction(undefined);`

karelhala (commented):
Looking good! Just one small thing and we are golden.

src/smart-components/common/page-action-route.js
Hyperkid123: already there!

karelhala (approved):

PR #556, opened 2020-10-26
Resolve splats in base role

PanSpagetka:
When base role permissions contains `*`and respected permissions are not at first page, no permission are preselected. 

This is starting to be crazy, needs to simplify the logic.

Issue 10: https://issues.redhat.com/browse/RHCLOUD-9983


Hyperkid123:
@PanSpagetka :crossed_swords:

Hyperkid123:
@PanSpagetka I think there is a commit that does not belong here.

src/redux/actions/permission-action.js
```diff
@@ -16,3 +16,25 @@ export const listPermissionOptions = ({ field, limit, offset, application, resou
   type: fieldToAction[field],
   payload: PermissionsHelper.listPermissionOptions(field, limit, offset, application, resourceType, verb, options),
 });
+
+export const expandSplats = ({
+  limit = 1000,
```
Hyperkid123, line 21: does this endpoint ignore the max API limit of 100?

PanSpagetka: When I tried it in API docs, it seems that 1000 is the limit.

Hyperkid123: Nice good to know :+1:

Hyperkid123 (approved):

PR #564, opened 2020-10-27
Fix 'resource definition' mobile styling in MUA

epwinchell:
This PR replaces the resource def button with a link in the MUA "compound expandable table," where the grid doesn't work nicely with buttons in mobile.
Jira: https://issues.redhat.com/browse/RHCLOUD-10034

Old
Screen Shot 2020-10-27 at 2 57 56 PM copy

New
Screen Shot 2020-10-27 at 2 55 58 PM


epwinchell:
cc @mmenestr

src/presentational-components/myUserAccess/ResourceDefinitionsLink.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+
+const ResourceDefinitionsLink = ({ onClick, access }) => {
+  return access.resourceDefinitions.length === 0 ? <span>N/A</span> : <Link to={(onClick = { onClick })}>{access.resourceDefinitions.length}</Link>;
```
Hyperkid123, line 6: I don't think the link onClick function will work. If you want to pass a function to react-router link it must return a pathname or route object with pathname `{pathname: 'foo'}`.

The link does accept `onClick` handler but it would be better for accessibility to give it `pathname` instead. You can also have both `onClick` to mutate the state and `to` to actually redirect the router.

mmenestr:
cc @beccaglass

beccaglass:
@epwinchell Styling fix looks good!

Can we include a description under the 'My User Access' title to clarify the different interaction pattern, as shown in the mocks? Also is it possible to make the table header sticky as well?

epwinchell:
`Can we include a description under the 'My User Access' title to clarify the different interaction pattern, as shown in the mocks? Also is it possible to make the table header sticky as well?`

@beccaglass Those changes were introduced here:  https://github.com/RedHatInsights/insights-rbac-ui/pull/490
94366271-18c37a80-00a5-11eb-8120-489ee144df61

beccaglass:
@epwinchell amazing thank you!!

epwinchell:
> @epwinchell amazing thank you!!

You're welcome. The only thing outstanding is a functional dropdown (the one above it a dummy) - it's in our queue.

Hyperkid123 (approved):

PR #571, opened 2020-11-03
Add info messages to add role table to show why some perms are missing

karelhala:
### Implements RHCLOUD-10392

When user copies role with unresolvable splats these permissions won't be shown in the table, so in order to avoid confusion users see info msg when copying a role

Screenshot from 2020-11-16 10-49-04

If user picks any role with unresolvable splats a popover is shown
Screenshot from 2020-11-16 10-48-57

And finally, if user filters down the list and nothing is found more descritive empty message is shown
Screenshot from 2020-11-16 10-48-37

### Recording
animation

### Note

I've added the unresolved apps so we could potentionally either show those permissions that can't be resolved or add them, but this is for more deep discussion on the JIRA issue.

### DIFF
--- a/src/smart-components/role/add-role-new/add-permissions-template.js
+++ b/src/smart-components/role/add-role-new/add-permissions-template.js
@@ -10,7 +10,7 @@ const AddPermissionTemplate = ({ formFields }) => {
   const [selectedPermissions, setSelectedPermissions] = useState(formOptions.getState().values['add-permissions-table'] || []);
 
   const unresolvedSplats =
-    formOptions.getState().values?.['copy-base-role']?.applications?.filter((app) => selectedPermissions?.find(({ uuid }) => uuid.includes(app))) ||
+    formOptions.getState().values?.['copy-base-role']?.applications?.filter((app) => !selectedPermissions?.find(({ uuid }) => uuid.includes(app))) ||
     [];
   const addPermissions = formFields[0][0];
   return (
--- a/src/smart-components/role/add-role-new/add-permissions-template.js
+++ b/src/smart-components/role/add-role-new/add-permissions-template.js
@@ -1,14 +1,17 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
-import { Chip, ChipGroup, Text, TextContent, Title } from '@patternfly/react-core';
+import { Chip, ChipGroup, Text, TextContent, Title, Button, Popover } from '@patternfly/react-core';
 import useFormApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-form-api';
-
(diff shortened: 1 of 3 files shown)


karelhala:
CC @mmenestr

Hyperkid123 (approved):
looks good. Let's wait for @mmenestr

john-dupuy:
Since `approval` does have granular permissions can we choose an app that does not in the examples, e.g. `patch`?

mmenestr:
Yeah, DON'T merge this yet :) Working with content person to review/finalize the wording!

karelhala:
@mmenestr texts has been updated.

src/smart-components/role/add-role-new/add-permissions-template.js
```diff
@@ -1,14 +1,17 @@
...
-
+import QuestionCircleIcon from '@patternfly/react-icons/dist/js/icons/outlined-question-circle-icon';
 import './add-role-wizard.scss';
 
 const AddPermissionTemplate = ({ formFields }) => {
   const formOptions = useFormApi();
   const [selectedPermissions, setSelectedPermissions] = useState(formOptions.getState().values['add-permissions-table'] || []);
 
+  const unresolvedSplats =
+    formOptions.getState().values?.['copy-base-role']?.applications?.filter((app) => selectedPermissions?.find(({ uuid }) => uuid.includes(app))) ||
```
PanSpagetka, line 13: I think that logic here should be negated. You are selecting resolved permissions. When i tried it patch has 1 resolved 1 unresolved so it seems to work. But for approval, that has all permissions resolved it also shows message.

PanSpagetka (changes requested):

src/smart-components/role/add-role-new/add-permissions-template.js
```diff
@@ -30,9 +33,32 @@ const AddPermissionTemplate = ({ formFields }) => {
         Add permissions
       </Title>
       <TextContent>
-        <Text>Select permissions to add to your role</Text>
+        <Text>
+          Select permissions to add to your role
+          {unresolvedSplats.length !== 0 && (
+            <Popover
+              headerContent="Custom roles only support granular permissions"
+              bodyContent="Wildcard permissions (for example, approval:*:*) aren’t included in this table and can’t be added to your custom role."
```
john-dupuy, line 41: We should use an app that doesn't have granular permissions here, e.g. `patch:*:*`

john-dupuy (approved):

src/smart-components/role/add-role-new/add-permissions-template.js
karelhala: Aaaah, yeah I've been messing around with the condition and pushed the wrong one. Good catch!

mmenestr:
This looks great, thank you!!

PanSpagetka (approved):

PR #574, opened 2020-11-04
Fix reloading base role when deleting all permissions

PanSpagetka:
This PR should be fixing the issue when base role permission are reloaded when you come to the Add permission step with no permissions selected.
https://issues.redhat.com/browse/RHCLOUD-10310
@john-dupuy


Hyperkid123 (commented):
code looks good. @PanSpagetka can you include a short video/giff of the flow?

PanSpagetka:
@Hyperkid123 yep, I have uploaded to the JIRA

Hyperkid123 (approved):

PR #575, opened 2020-11-04
Add role wizard review states.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-10299

### Changes
- added wizard context with submit metadata
- added add role success review state
- added add role error review
- removed the success notification (i know its in the giff)
add-role-statuses

### DIFF
--- a/src/smart-components/role/add-role-new/review-step-buttons.js
+++ b/src/smart-components/role/add-role-new/review-step-buttons.js
@@ -1,0 +1,75 @@
+import { Button } from '@patternfly/react-core';
+import React, { Fragment, useContext } from 'react';
+import PropTypes from 'prop-types';
+import get from 'lodash/get';
+import { AddRoleWizardContext } from './add-role-wizard';
+import useFormApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-form-api';
+
+const selectNext = (nextStep, getState) =>
+  ({
+    string: nextStep,
+    function: nextStep({ values: getState().values }),
+  }?.[typeof nextStep] || nextStep.stepMapper?.[get(getState().values, nextStep.when)]);
+
+const NextButton = ({ nextStep, valid, handleNext, nextLabel, getState, handleSubmit, submitLabel }) => (
+  <Button
+    variant="primary"
+    type="button"
+    isDisabled={!valid || getState().validating}
+    onClick={() => (nextStep ? handleNext(selectNext(nextStep, getState)) : handleSubmit())}
+  >
+    {nextStep ? nextLabel : submitLabel}
+  </Button>
+);
+
+NextButton.propTypes = {
+  nextStep: PropTypes.oneOfType([PropTypes.string, PropTypes.func, PropTypes.object]),
+  handleSubmit: PropTypes.func.isRequired,
+  submitLabel: PropTypes.node.isRequired,
+  valid: PropTypes.bool,
+  handleNext: PropTypes.func.isRequired,
+  nextLabel: PropTypes.node.isRequired,
+  getState: PropTypes.func.isRequired,
+};
+
+const ReviewStepButtons = ({ disableBack, handlePrev, nextStep, handleNext, buttonLabels: { cancel, submit, back, next } }) => {
+  const formOptions = useFormApi();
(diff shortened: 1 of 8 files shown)


john-dupuy (approved):
LGTM, where does the "Add role to group" button take you? The groups page?

Hyperkid123:
> LGTM, where does the "Add role to group" button take you? The groups page?

Yes. Currently, there is no screen that would let you select a user and assign them to a group. So that is the only logical place we can redirect at the moment.

PanSpagetka (approved):

sitemap.xml
```diff
@@ -1,26 +1,26 @@
...
-<url> <loc>https://cloud.redhat.com/roles/edit/:id</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:uuid</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:id/remove</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:id/edit</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:roleId/permission/:permissionId</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:roleId/permission/:permissionId/edit</loc> </url>
-<url> <loc>https://cloud.redhat.com/users</loc> </url>
-<url> <loc>https://cloud.redhat.com/users/detail/:username</loc> </url>
-<url> <loc>https://cloud.redhat.com/my-user-access</loc> </url>
+<url> <loc>https://cloud.redhat.com/undefined/undefined/</loc> </url>
```
karelhala, line 3: The plugin is probably broken @PanSpagetka can you look at it? While this is being resolved, no need to push this file.

src/smart-components/role/add-role-new/review-step-buttons.js
```diff
@@ -0,0 +1,83 @@
+import { Button } from '@patternfly/react-core';
+import React, { Fragment, useContext } from 'react';
+import PropTypes from 'prop-types';
+import get from 'lodash/get';
+import { AddRoleWizardContext } from './add-role-wizard';
+import useFormApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-form-api';
+
+const selectNext = (nextStep, getState) => {
+  if (typeof nextStep === 'string') {
```
karelhala, line 9: This branching looks a bit complicated, how about object mapper?
```JS
const selectNext = (nextStep, getState) => ({
  string: nextStep,
  function: nextStep({ values: getState().values })
})?.[typeof nextStep] || nextStep.stepMapper?.[get(getState().values, nextStep.when)]
```

src/smart-components/role/add-role-new/review-step-buttons.js
```diff
@@ -0,0 +1,83 @@
...
+  handleNext: PropTypes.func.isRequired,
+  nextLabel: PropTypes.node.isRequired,
+  getState: PropTypes.func.isRequired,
+};
+
+const ReviewStepButtons = ({ disableBack, handlePrev, nextStep, handleNext, buttonLabels: { cancel, submit, back, next } }) => {
+  const formOptions = useFormApi();
+  const { success, error, submitting } = useContext(AddRoleWizardContext);
+  if (success || typeof error === 'undefined' || error || submitting) {
+    return null;
```
karelhala, line 47: Wouldn't it be better to return `<Fragment />` here? So we are consistent on return type.

karelhala (commented):
I really like the usage of context here!

src/smart-components/role/add-role-new/review-step-buttons.js
Hyperkid123: It does not really matter. The fragment will only show up in the virtual DOM but the effect will be the same on the actual DOM. And this way there is one less node to traverse in the react DOM with null.

sitemap.xml
PanSpagetka: Yep, I think that we can delete the plugin. The sitemap is kind of useless anyway because all the links are hidden behind authentication.

PanSpagetka: We can move discussion related to sitemap here: https://github.com/RedHatInsights/frontend-components/pull/804

karelhala (approved):

PR #576, opened 2020-11-04
Fix add role permissions chips filtering.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-10493

### Changes
- use checkbox filter instead of group filter (they are the same as far as I can tell and the checkbox actually works)

add-role-permi


john-dupuy:
Nice @Hyperkid123! Can we also do the same for the filter on the role details page?

Hyperkid123:
Ah I see that is broken as well. Sure thing.

Hyperkid123:
I see this is almost copy/paste. I think it would deserve refactoring these components (different task).

john-dupuy (approved):
Thanks @Hyperkid123 :rocket:

sitemap.xml
```diff
@@ -1,26 +1,26 @@
...
-<url> <loc>https://cloud.redhat.com/roles/edit/:id</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:uuid</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:id/remove</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:id/edit</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:roleId/permission/:permissionId</loc> </url>
-<url> <loc>https://cloud.redhat.com/roles/detail/:roleId/permission/:permissionId/edit</loc> </url>
-<url> <loc>https://cloud.redhat.com/users</loc> </url>
-<url> <loc>https://cloud.redhat.com/users/detail/:username</loc> </url>
-<url> <loc>https://cloud.redhat.com/my-user-access</loc> </url>
+<url> <loc>https://cloud.redhat.com/undefined/undefined/</loc> </url>
```
karelhala, line 3: No need to push this file.

karelhala (commented):
Looking good! The Group filter was probably used because previously there was `show more` option which is no longer in use.

sitemap.xml
Hyperkid123: Oh yeah... It's generated when you run install.

PR #581, opened 2020-11-05
Display name sorting

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-10219

We have an incredible amount of copy-paste code when it comes to roles tables. We should merge them at some point.

Can't add display_name sorting to tables that are using `/groups/<uuid>/roles` [endpoint](https://github.com/RedHatInsights/insights-rbac/blob/master/docs/source/specs/openapi.json#L826) because it does not support it.

### DIFF
--- a/src/smart-components/group/add-group/roles-list.js
+++ b/src/smart-components/group/add-group/roles-list.js
@@ -1,14 +1,13 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
+import { sortable } from '@patternfly/react-table';
 import { mappedProps } from '../../../helpers/shared/helpers';
 import { defaultCompactSettings, defaultSettings } from '../../../helpers/shared/pagination';
 import { TableToolbarView } from '../../../presentational-components/shared/table-toolbar-view';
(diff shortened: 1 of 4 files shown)


src/smart-components/group/add-group/roles-list.js
```diff
@@ -26,11 +25,15 @@ const createRows = (data, expanded, checkedRows = []) => {
     : [];
 };
 
-const RolesList = ({ roles, fetchRoles, isLoading, pagination, selectedRoles, setSelectedRoles }) => {
+const RolesList = ({ roles, fetchRoles, isLoading, pagination, selectedRoles, canSort, setSelectedRoles }) => {
   const [filterValue, setFilterValue] = useState('');
+  const { current: columns } = useRef([
```
karelhala, line 30: I think you don't want to recalculate these rows on each render right? Wouldn't it be better to use `useCallback`?
```JS
const columns = useCallback(() => ([
  { title: 'Name', key: 'display_name', ...(canSort ? { transforms: [sortable] } : { orderBy: 'name' }) },
  { title: 'Description' },
]),[ canSort ])();

Hyperkid123: Yup, that is why I am using the ref. It's static and you don't need any function calls.

karelhala (approved):

PR #582, opened 2020-11-05
Disable row and show tooltip for cost permissions if no resource available

PanSpagetka:
Disable row and show tooltip for cost permissions that doesn't have any resources defined.

https://issues.redhat.com/browse/RHCLOUD-9983
@mmenestr
Screenshot from 2020-11-20 12-27-36

### DIFF
--- a/src/smart-components/role/add-role-new/DisabledRowWrapper.js
+++ b/src/smart-components/role/add-role-new/DisabledRowWrapper.js
@@ -8,8 +8,8 @@ export const DisabledRowWrapper = ({ row, ...props }) =>
     <Tooltip
       content={
         <div>
-          To add this permission to your role and define specific resources for it, at least one data source must be connected. Go{' '}
-          <a href="https://cloud.redhat.com/settings/sources">here</a> to add a new Source for Cost Management.
+          To add this permission to your role and define specific resources for it, at least one data source must be connected.{' '}
+          <a href="./settings/sources">Configure sources for Cost Management</a>
         </div>
       }
       exitDelay={2000}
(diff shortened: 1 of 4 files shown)


mmenestr:
Can you send me a screenshot of what this looks like? Or add it here?

PanSpagetka:
@mmenestr yep, sorry I forgot to include it initially

mmenestr:
Ok don't merge this yet because I'd like to run it by Nicole! I'm thinking there might be a different way to show this

mmenestr:
@PanSpagetka would it be possible to show that message inside the dropdown - almost as an empty state, when there are no resources to choose from?

PanSpagetka:
@mmenestr I have updated code and screenshot to match changes we talked about.

Hyperkid123 (approved):

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -293,6 +308,23 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...p
           },
         ]}
         isFilterable={true}
+        rowWrapper={({ row, ...props }) =>
```
karelhala, line 311: Can you add this to custom component? If this would have been something small I would ok with it being inline, but since you added tooltip and such I'd much more happy if this would have been in special file.

src/smart-components/role/add-role-new/DisabledRowWrapper.js
```diff
@@ -0,0 +1,26 @@
...
+import { Tooltip } from '@patternfly/react-core';
+import { RowWrapper } from '@patternfly/react-table';
+
+export const DisabledRowWrapper = ({ row, ...props }) =>
+  row.disableSelection ? (
+    <Tooltip
+      content={
+        <div>
+          To add this permission to your role and define specific resources for it, at least one data source must be connected. Go{' '}
+          <a href="https://cloud.redhat.com/settings/sources">here</a> to add a new Source for Cost Management.
```
karelhala, line 12: Can you change this URL to be env agnostic?
```suggestion
          <a href="./settings/sources">here</a> to add a new Source for Cost Management.
```

karelhala (approved):

PR #583, opened 2020-11-06
Fix group filter chips deletion.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-10493

### Changes
- reverts: https://github.com/RedHatInsights/insights-rbac-ui/pull/576
- fixes incorrect chips deletion for toolbar group filter

groups-delete-fix


john-dupuy (approved):
Thanks @Hyperkid123, checked this out locally and it looks good to me!

PR #584, opened 2020-11-09
Use display name in group roles and mua roles tables.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-10219

@john-dupuy I hope these are the last lines.


karelhala (approved):

john-dupuy:
@Hyperkid123 from what I can tell from the code, it looks like `display_name` needs to be added to the `fetchRolesForGroup` helper function? I am still seeing the `name` used on the group details page in qa-beta/ci-beta. 

https://github.com/Hyperkid123/insights-rbac-ui/blob/df65a67dff9307ed6a92619fee9a6f541142ccc3/src/helpers/group/group-helper.js#L51-L53

Hyperkid123:
@john-dupuy that is the endpoints where we use it yet `/groups/<id>/roles`.

Hyperkid123:
There are some for the groups though. This is a nightmare. We have so many duplicate components that do the same thing it's making this very annoying.

PR #587, opened 2020-11-18
Group detail page kebab & inconsistencies

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-9499

_1. When editing the name/description of a group, the primary button says "Submit", but when editing the name/description of a role, the primary button says "Save". Since role editing was more recent, I'm guessing "Save" is the proper title here._ 
**Seems to be already fixed**

_2. When editing a group (either by changing name/description or adding roles/members),  a success notification appears when changes are successful. When the user cancels the edit, a warning notification appears. However, no such notification appears for custom roles._
**Added the notification**

**...also added missing failure notification for removing a role**

_...3. and 4. already fixed/not issues_

_5. Kebab on the details page of custom roles should also be on the details page of custom groups_
**Added the kebab with edit & remove actions**

_6. When removing more than 1 permission at a time, only the first permission is displayed in the modal text. According to the mocks, the number of permissions should be shown._ 
**Fixed in:** 
https://github.com/RedHatInsights/insights-rbac-ui/pull/482/files

@mmenestr

PR #590, opened 2020-11-18
Fix ouia ID on group dropdown

john-dupuy:
cc @fhlavac I missed this on https://github.com/RedHatInsights/insights-rbac-ui/pull/587


fhlavac:
@john-dupuy oh, thanks for fixing that

fhlavac (approved):

PR #605, opened 2020-12-02
Dont select disabled rows in Add permission step

PanSpagetka:
Don't select disabled rows when creating new role based on existing cost role.

https://issues.redhat.com/browse/RHCLOUD-10915

@john-dupuy


karelhala (approved):
I don't really like two filters, but since this is fairly complex filtering I'll give it a pass.

PanSpagetka:
> I don't really like two filters, but since this is fairly complex filtering I'll give it a pass.

I could make it one, but it seemed more readable to have it split in two.

Hyperkid123:
Readability ftw in this case.

PR #627, opened 2020-12-17
Hide add permission button in non beta and for system roles.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-10977
fixes: https://issues.redhat.com/browse/RHCLOUD-11126

### Changes
- hide the add permission in non beta env
- hide the add permission button for system roles

It is missing from acceptance criteria but belongs to custom roles stuff (thanks @john-dupuy). Waiting for confirmation that we should also hide this behind the beta flag.

### DIFF
--- a/src/smart-components/role/role-permissions.js
+++ b/src/smart-components/role/role-permissions.js
@@ -141,7 +141,7 @@ const Permissions = () => {
   ];
 
   const toolbarButtons = () => [
-    window.insights.chrome.isBeta()
+    ...(window.insights.chrome.isBeta() && !role.system
       ? [
           // eslint-disable-next-line react/jsx-key
           <Link to={`/roles/detail/${role.uuid}/role-add-permission`} key="role-add-permission" className="pf-m-visible-on-md">
@@ -149,17 +149,17 @@ const Permissions = () => {
               Add Permission
             </Button>
           </Link>,
+          {
+            label: 'Add Permission',
+            props: {
+              className: 'pf-m-hidden-on-md',
+            },
+            onClick: () => {
+              history.push(`/roles/detail/${role.uuid}/role-add-permission`);
+            },
+          },
         ]
-      : [],
-    {
-      label: 'Add Permission',
-      props: {
-        className: 'pf-m-hidden-on-md',
-      },
-      onClick: () => {
-        history.push(`/roles/detail/${role.uuid}/role-add-permission`);
-      },
-    },
+      : []),
     {
       label: 'Remove',
       props: {
--- a/src/smart-components/role/role-permissions.js
+++ b/src/smart-components/role/role-permissions.js
@@ -140,10 +140,9 @@ const Permissions = () => {
     },
   ];
 
-  const toolbarButtons = () => [
-    ...(window.insights.chrome.isBeta() && !role.system
+  const toolbarButtons = () =>
+    window.insights.chrome.isBeta() && !role.system
       ? [
-          // eslint-disable-next-line react/jsx-key
           <Link to={`/roles/detail/${role.uuid}/role-add-permission`} key="role-add-permission" className="pf-m-visible-on-md">
             <Button variant="primary" aria-label="Add Permission">
               Add Permission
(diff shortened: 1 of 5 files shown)


src/smart-components/role/role-permissions.js
```diff
@@ -141,45 +141,47 @@ const Permissions = () => {
...
-              confirmButtonLabel: multiplePermissionsSelected ? 'Remove permissions' : 'Remove permission',
-            },
-          });
-        },
+      onClick: () => {
+        history.push(`/roles/detail/${role.uuid}/role-add-permission`);
+      },
+    },
+    {
+      label: 'Remove',
```
john-dupuy, line 164: Is this part also gated by `beta`? If not it should be.

Hyperkid123: No only the first item in the array.

Hyperkid123: It's still visible on small screens! Nice catch.

john-dupuy: Oh interesting, for some reason the dropdown is hidden on large screens.

Hyperkid123: It's a bit weird. I think I'll have to look into that component. Because if you now have both the add permission and remove, you can't remove the selected permission on a large screen because the dropdown is just not there...

Hyperkid123: So it won't affect prod but in prod beta users won't be able to remove permissions. Here's a question, should they be even able to remove permissions in prod? They can't add any so I would find it strange if users could remove them.

john-dupuy: Yeah, this is a part of https://issues.redhat.com/browse/RHCLOUD-11126 btw

john-dupuy: I think the gating for the "Remove" dropdown item should be the same as for the "Add permissions" button:
* only present for non-system roles in beta environments

Hyperkid123: Right. It looks like missing CSS. The element is in the DOM. Only it has a `display: none` CSS property. After fixing it in the browser console it shows properly. I think I will try and fix it on Monday. And before we merge to prod.

> I think the gating for the "Remove" dropdown item should be the same as for the "Add permissions" button

Agreed.

Hyperkid123:
@john-dupuy alright. The buttons are now properly hidden and the dropdown is now visible.

john-dupuy:
@Hyperkid123 thanks for this, checking this PR locally I noticed that in the `beta` environment on the roles page that the dropdown has no items in it. Screenshot below.
Screenshot from 2021-01-04 08-41-47

Additionally, I noticed that in non-beta environments - users are still able to edit and delete custom roles via the dropdown present on a row within the table. Similarly on the role details page, a user is able to individually remove permissions using this dropdown. Do we want to also gate these dropdowns?

Hyperkid123:
> the dropdown has no items in it
This will be a bug in FEC. I bet when you shrink the window the dropdown will be populated with that one item (Create role). We need to fix it there.

>  users are still able to edit and delete custom roles via the dropdown present on a row within the table
We gate those as well. (I hate feature flags)

Hyperkid123:
@john-dupuy 
> users are still able to edit and delete custom roles via the dropdown present on a row within the table

On second thought. If users can create custom roles, there should be no custom roles in the table so I don't really see why to gate these options. (Yes I know the API is avaiable)

john-dupuy:
> I bet when you shrink the window the dropdown will be populated with that one item (Create role).

Yep, just confirmed this was the case. I guess eventually we could put a "Remove" button there too? Provided that we also have a way to select multiple custom roles.

Hyperkid123:
I would assume so. I will try and fix it inside FEC, but I would not block this by that. This issue present in each app, but nobody noticed it probably.

john-dupuy:
> If users can create custom roles, there should be no custom roles in the table so I don't really see why to gate these options

Yeah that's a good point. Custom roles could be in the table if a user makes them in a beta environment however and goes back to the non-beta environment. We might hit an edge case where someone removes the custom role accidentally in the non-beta environment and is confused as to why they can't create a new one.

Hyperkid123:
@john-dupuy the empty dropdown is now fixed. I have removed some CSS rules which were hiding the elements. So the dropdown was not actually empty, but the elements were hidden. So it's fortunately not FEC problem.

It's now using the dedicated action instead which was created for this exact use-case.

dropdown

john-dupuy (approved):

PR #629, opened 2021-01-05
Clear filter on delete

PanSpagetka:
When you delete group from detail, you are redirected to the `/roles` or `/groups` list with cleared filters. 
https://issues.redhat.com/browse/RHCLOUD-10569

### DIFF
--- a/src/smart-components/group/group.js
+++ b/src/smart-components/group/group.js
@@ -87,13 +87,13 @@ const Group = ({
     <DropdownItem
       component={
         <Link
-          to={() => {
-            onDelete();
-            return (location.pathname.includes('members') ? routes['group-detail-members-remove'] : routes['group-detail-roles-remove']).replace(
+          onClick={onDelete}
+          to={() =>
+            (location.pathname.includes('members') ? routes['group-detail-members-remove'] : routes['group-detail-roles-remove']).replace(
               ':uuid',
               uuid
-            );
-          }}
+            )
+          }
         >
           Delete
         </Link>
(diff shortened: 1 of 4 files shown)


src/smart-components/group/group.js
```diff
@@ -86,10 +87,13 @@ const Group = ({
     <DropdownItem
       component={
         <Link
-          to={(location.pathname.includes('members') ? routes['group-detail-members-remove'] : routes['group-detail-roles-remove']).replace(
-            ':uuid',
-            uuid
-          )}
+          to={() => {
+            onDelete();
```
Hyperkid123, line 91: @PanSpagetka did you try hooking the onDelete callback to `onClick` event? That might be the "correct" way of doing it. If it works.

src/smart-components/role/role.js
```diff
@@ -42,7 +43,16 @@ const Role = () => {
   const dropdownItems = [
     <DropdownItem component={<Link to={routes['role-detail-edit'].replace(':id', uuid)}>Edit</Link>} key="edit-role" />,
     <DropdownItem
-      component={<Link to={routes['role-detail-remove'].replace(':id', uuid)}>Delete</Link>}
+      component={
+        <Link
+          to={() => {
+            onDelete();
```
Hyperkid123, line 49: ^^

src/smart-components/group/group.js
PanSpagetka: Yep, that's better. For some reason I thought it doesn't work, but actually it does.

Hyperkid123 (approved):

PR #626, opened 2020-12-17
Sync prod stable with latest changes.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-10977

wait for: https://github.com/RedHatInsights/insights-rbac-ui/pull/627

PR #636, opened 2021-01-11
Sync prod-stable with master

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-11346

Replaces: https://github.com/RedHatInsights/insights-rbac-ui/pull/626 because I've made a booboo and merged it instead of prod-beta sync. And then had to revert the changes :facepalm:


Hyperkid123:
@john-dupuy Alright. This PR should be prod stable ready with all the bug fixes. MUA stuff should be also included in the RBAC UI.

Chrome and navigation change PRs are out but waiting for RBAC changes first.

john-dupuy (approved):
Tested this PR out locally and it LGTM. I do think it is odd that editing/deleting custom roles is possible in non-beta, but creating roles is locked to beta. But I understand why this is the case and don't think many people will hit these scenarios.

ryelo (approved):

PR #639, opened 2021-01-19
Set rbac/users as defualt url for org admins

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-11594

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -21,7 +21,7 @@ InsightsRoute.propTypes = {
   rootClass: PropTypes.string,
 };
 
-export const Routes = () => {
+export const Routes = ({ isOrgAdmin }) => {
   return (
     <Suspense fallback={<AppPlaceholder />}>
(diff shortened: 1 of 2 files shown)


src/routes.js
```diff
@@ -21,7 +21,7 @@ InsightsRoute.propTypes = {
   rootClass: PropTypes.string,
 };
 
-export const Routes = () => {
+export const Routes = (isOrgAdmin) => {
```
karelhala, line 24: Wrong consumption of props
```suggestion
export const Routes = ({ isOrgAdmin }) => {
```

karelhala (approved):

ryelo:
Leaving a comment here too, I'm not sure that this PR does exactly what we're looking for. 

MUA should be the default route if someone goes to /settings

If someone goes to /settings/rbac, we should redirect them to /settings/rbac/users

PR #640, opened 2021-01-19
Set Group and Role name max length limit = 150

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-11440

### DIFF
--- a/src/smart-components/group/add-group/group-information.js
+++ b/src/smart-components/group/add-group/group-information.js
@@ -18,7 +18,7 @@ const GroupInformation = (formValue, onHandleChange, setIsGroupInfoValid, isGrou
             fieldId="group-name"
             helperTextInvalid={
               formValue.name?.trim().length > 0 && !isValidating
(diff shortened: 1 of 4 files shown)


src/smart-components/group/add-group/group-information.js
```diff
@@ -18,7 +18,9 @@ const GroupInformation = (formValue, onHandleChange, setIsGroupInfoValid, isGrou
             fieldId="group-name"
             helperTextInvalid={
               formValue.name?.trim().length > 0 && !isValidating
-                ? 'This group name already exists. Please input a unique group name.'
+                ? formValue.name?.trim().length < 150
```
Hyperkid123, line 21: I don't think we should trim the length.  I think you can crash the API even with 150 spaces. Try sending something like 
```
a                                                      b                                            c                               #actual end of the string
```

Hyperkid123 (approved):

PR #643, opened 2021-01-19
use patch roles instead of update

ryelo:
RHCLOUD-10391

This should allow users to edit a role immediately after creating it. 

The only weird thing I've found with this so far is that I don't think the modal itself is using the display_name. I also am not sure if it's okay that we set the name and display_name to the same value, but as far as I know that shouldn't break anything.


src/redux/actions/role-actions.js
```diff
@@ -93,6 +93,29 @@ export const updateRole = (roleId, data, useCustomAccess) => ({
...
+  meta: {
+    notifications: {
+      fulfilled: {
+        variant: 'success',
+        title: 'Success updating role',
+        dismissDelay: 8000,
+        dismissable: false,
+        description: 'The role was updated successfully.',
+      },
+      rejected: {
```
Hyperkid123, line 108: I would remove the rejected key. The default error notification will display the API error message.

john-dupuy (approved):
LGTM, tried this out locally and I was able to successfully edit the role without first navigating to the details page of the role.

Hyperkid123 (approved):

PR #646, opened 2021-01-20
Utils v3

Hyperkid123:
Brings in the new utils build from: https://github.com/RedHatInsights/frontend-components/pull/897

~~Also depends on: https://github.com/RedHatInsights/insights-rbac-ui/pull/638~~

Introduces about 50% size reduction for the utils package. Which is quite nice.

### Before
Screenshot from 2021-01-20 14-51-56

### After
Screenshot from 2021-01-20 14-45-19

Now the big bad boy of utils package is actually the Axios. Actual FEC source code has 4kb only.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4281,9 +4281,9 @@
       }
     },
     "@redhat-cloud-services/frontend-components-utilities": {
-      "version": "2.2.9",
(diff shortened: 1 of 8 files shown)


PanSpagetka (approved):
I did some clicking in UI and it seems that nothing is broken :+1:

PR #647, opened 2021-01-21
Change base path for RBAC and update links

PanSpagetka:
Use `/settings` as a base path for MUA and `settings/rbac` as a base path for RBAC.

`/settings` -> `/settings/my-user-access` this redirect is done in `App.js:23` no need to have it as default route. (Hopefully, for some reason can't test this locally.)
`/settings/rbac` -> `/settings/rbac/users` this will be new default route.


src/AppEntry.js
```diff
@@ -7,7 +7,7 @@ import { getBaseName } from '@redhat-cloud-services/frontend-components-utilitie
 
 const InsightsRbac = () => (
   <Provider store={store}>
-    <Router basename={getBaseName(location.pathname)}>
+    <Router basename={getBaseName(location.pathname, 2).includes('rbac') ? '/settings/rbac' : 'settings'}>
```
Hyperkid123, line 10: I think the `settings` basename should also include `/`.

(Not 100% sure on this one) You also have to take beta into consideration. I think the `getBaseName` should include the `beta` partial if its necessary.

PanSpagetka: Yep, I wanted to make it more explicit, but `getBaseName` should handle these cases, so it will be better to use it.

john-dupuy (approved):
Testing this locally:
- From the landing page, clicking the settings cog goes to MUA
- Using the app filter, clicking "User Access" goes to User Access -> Users
So this appears to be working as expected. 
Only thing is that if I put only "settings" in the URL (no RBAC) I was met with an infinite load. I think this is what you mentioned by not being able to test that part locally however.

PR #651, opened 2021-01-22
[DO NOT MERGE]Pf dependencies debugging

Hyperkid123:
This PR shows that for some reason, referencing pf-react-table causes the whole PF react-core to be injected into the production build.

How to reproduce:

1. install dependencies
2. run `yarn analyze`, you will see a graph in browser, the whole bundle will have 784.2 KB
3. go here and use PF react-table import instead of const override: https://github.com/RedHatInsights/insights-rbac-ui/pull/651/files#diff-0f4d89b4ef9387451ee47bbf2648ab8a4eba081195d459390f68842902738ea4R31
4. run `yarn analyze`, you will see a graph in browser, the whole bundle will have 2.05 MB and all PF components are in the bundle. Search for about modal or login page. It will be present in the graph even though they are never used in rbac UI.

cc @RedHatInsights/platform-experience @karelhala @redallen

@evwilkin I know that you are aware I just wanted to show a real example as well.

### Witouth PF
screenshot-127 0 0 1_8888-2021 01 22-14_47_55

### With PF
screenshot-127 0 0 1_8888-2021 01 22-14_48_17


redallen:
The reason is this line in https://github.com/RedHatInsights/insights-rbac-ui/blob/master/config/federation-plugin.js#L20:
```js
{ '@patternfly/react-core': { singleton: true, requiredVersion: dependencies['@patternfly/react-core'] } },
```

Seeing that any part of `@patternfly/react-core` is in the bundle, Webpack prepares itself to federate the whole thing. Simply remove that line and you'll only get the parts of `react-core` that `react-table` depends on.

Hyperkid123:
> Simply remove that line and you'll only get the parts of react-core that react-table depends on.

@redallen I don't think it's simple as that. Removing PF from dependency sharing is not something we want to do. Tree shaking should work with federated modules. We also know that other libraries using referencing PF react-core (like the FEC or data-driven forms) do not cause this issue.

PF is by far the largest dependency in all apps and if we should share any module, it should be the react-core library. It is actually more beneficial to leave it "broken". After as little as 4 app switches in chrome SPA, the whole react-core shared module will be smaller than importing a whole new react-core vendor module from a new app.
 
Surely there is a better solution than turning off the most significant shared module.

redallen:
Try setting your `@patternfly/react-core` version to be `^4.89.1` to match `@patternfly/react-table`'s. Tthen use `import { cellWidth } from '@patternfly/react-core/dist/esm/components/Table/utils/decorators/cellWidth';`.You have to use the long import until https://github.com/patternfly/patternfly-react/pull/5366 lands because all the decorators are in the same file which ends up including most of react-table.

With these fixes you should have a 550Kb pfVendor with only part of react-core.

Hyperkid123:
@redallen nice I will try it out later today. Also, I will check if other modules from the table could cause the issue.

Hyperkid123:
OK, I can confirm that the direct import path does work. I will now try the rest of the commonly used modules.

Hyperkid123:
@redallen Unfortunately it seems that is the issue is common for all of the table components. I have used the direct import to the decorator, however I have then tried directly importing the table component and the whole react-core is back in the bundle:
```jsx
import { Table } from '@patternfly/react-table/dist/esm/components/Table/Table';
console.log({ Table });
```
I have pushed the change to this PR.

screenshot-127 0 0 1_8888-2021 01 26-14_58_43


EDIT: I have gone through the actual ESM dependency and I see that there is a mix of direct/relative imports from `react-core` after commenting out relative imports, the bundle only contained react-core components that were actually used.

redallen:
Yes, see https://github.com/patternfly/patternfly-react/pull/4906.

PR #654, opened 2021-01-26
Update frontend components to v3

Hyperkid123:
Another FEC package update. In this repo, v3 reduced the size of `@redhat-cloud-services/frontend-components` module in the production bundle by about  60%

depends on #646 

### Before
Screenshot from 2021-01-26 13-46-15

### After
Screenshot from 2021-01-26 13-45-07


karelhala (approved):

Hyperkid123:
@karelhala I'll update the version to stable v3 and fix snapshots

PR #655, opened 2021-01-26
Validate role/group dscription length

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-9804

### DIFF
--- a/src/smart-components/role/add-role-new/schema.js
+++ b/src/smart-components/role/add-role-new/schema.js
@@ -66,6 +66,12 @@ export default (container) => ({
                 when: 'role-type',
                 is: 'create',
               },
+              validate: [
+                {
+                  type: 'max-length',
+                  threshold: 150,
+                },
+              ],
             },
             {
               component: 'base-role-table',
@@ -129,6 +135,12 @@ export default (container) => ({
               name: 'role-copy-description',
               type: 'text',
               label: 'Role description',
+              validate: [
+                {
+                  type: 'max-length',
+                  threshold: 150,
+                },
+              ],
             },
           ],
         },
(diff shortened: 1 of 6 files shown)


john-dupuy (changes requested):
Checking this out locally I was able to verify that the name was checked and I was unable to click submit.

However, I enter in a really long name, RBAC will still crash -- I think because it is trying to make an API request to see if the name exists? If the text is over 150 characters, we shouldn't check that. 

I believe this is the behavior for groups, as I can't crash RBAC in a similar way when editing the name of a group.

Hyperkid123:
> However, I enter in a really long name, RBAC will still crash -- I think because it is trying to make an API request to see if the name exists? If the text is over 150 characters, we shouldn't check that.

That would be the name async validation. We have to limit that aswell. Probably just don't trigger the call.

PanSpagetka:
After last commit it shouldn't fire request if group name is too long

john-dupuy:
@PanSpagetka ah, I meant we need to do that for roles, groups are already working properly!

john-dupuy (approved):
Testing out locally, I can no longer crash RBAC by putting in a long role name in the edit modal. Nice work @PanSpagetka :+1:

src/smart-components/role/add-role-new/schema.js
```diff
@@ -66,6 +66,12 @@ export default (container) => ({
                 when: 'role-type',
                 is: 'create',
               },
+              validate: [
+                {
+                  type: 'max-length',
```
Hyperkid123, line 71: Can you use the constant as in the edit group modal?

PR #656, opened 2021-01-26
Fix path to notification portal

Hyperkid123:
After one of the updates for notifications package which adds backwards compatibility to older FEC packages, there is a conflict between file path and folder path (they have the same name). And that causes the component to be `undefined` because its referencing the wrong asset (v2 asset).

Appending the path with `/` will resolve the folder (which we want) as the source of the asset.


john-dupuy (approved):

Hyperkid123:
I should add that the compatibility assets will exist only until all is migrated to v3.

PR #665, opened 2021-01-28
Jira 10596 - Pending Changes

arivepr:
Have gotten through most of the card, specific issues are shown here in hopes to find a fix for them. 
Here's the original list of issues: 
Screen Shot 2021-01-28 at 10 09 55 AM

### Issues that have been completed thus far: 

**1.** wip
**2.** If a role has no permissions,  there is no empty state.
Screen Shot 2021-01-28 at 10 34 36 AM
**3.** Labels should appear when permissions are selected.
Screen Shot 2021-01-28 at 10 39 49 AM
**4.** Exit warning should only appear when having made changes --- fixed, reproducible at add-permission-wizard.
**5.** wip
**6.** Success notification when permissions are successful. --- Added with a caveat* (also, it's imitating create role wizard)
Screen Shot 2021-01-28 at 10 47 13 AM
**7.** The page should reload after adding permissions to refresh the permission in role --- fixed reloading behavior, caveat**
**8.** No way to add cost resource definitions when selecting cost permissions --- wip.
**Issues 9-12 on card should all be fixed (uppercase words), please notify if otherwise.**

### Caveats -- Help Needed!
**6.** I'm getting an error from the _updateRole_ action helper whenever I try to save the added permissions to the role. 
Screen Shot 2021-01-28 at 11 12 04 AM
**Here's what I'm sending the api:**
Screen Shot 2021-01-28 at 11 16 05 AM

*the api call is happening in add-role-permission-view

**7.** I changed the behavior to reload the page, which should give updated data, but cannot test this out without the data actually being saved through the api.


mmenestr:
2. There are spelling mistakes in the body + the button is too close to the body. Are you using the [Patternfly Empty state component](https://www.patternfly.org/v4/components/empty-state#basic)? All that spacing should be in there by default!
6. Empty state should be centered on the screen because it's in a wizard!

PanSpagetka (approved):
All issues except no. 5 should be fixed.

mmenestr:
@PanSpagetka @fhlavac Just reviewed all this in ci beta and found some issues:

1. there's a typo in this empty state still - should be "one" permission, not "on" permission. Also the header looks really small - should be 20px not 16px font!

image

2. Slightly different thing but in the review page of the "add permission" wizard, I would change the label from "Adding permissions" to "**Added permissions**", and remove the ":" after each "heading" because we don't usually use those in the other wizard review pages...

image

3. Last thing is that when I add a permission, the success toast alert blocks the wizard "X" button - just kind of a weird interaction... I would expect to see the toast once the user completely exited the wizard. OR just have one or the other, either an exit success screen OR a success toast.... having both seems redundant now that i've tried it out in ci beta.
image

@john-dupuy for thoughts :)

PR #667, opened 2021-02-01
Fix calculateChecked to allow unselect all groups using BulkSelect

fhlavac:
Fixes https://issues.redhat.com/browse/RHCLOUD-11127

CalculateChecked helper function returned a wrong value since the function was assuming that "Select all" selects all the rows but "Custom default access" group is never selected.

@john-dupuy


john-dupuy (approved):
Awesome! Tested this out locally and can confirm it fixes the issue. Thanks @fhlavac :+1:

Hyperkid123:
@fhlavac I am unable to update the branch from GH UI. Can you rebase please?

PR #670, opened 2021-02-03
Use composed filters when making sort requests in TableToolbar

PanSpagetka:
It worked for simple filters where you are filtering by name (Roles,Groups) but when there was multiple different filters present. Sort request didn't reflect that. Now onSort request are using same logic as onFIliter requests.

https://issues.redhat.com/browse/RHCLOUD-11475
https://issues.redhat.com/browse/RHCLOUD-11553


john-dupuy (approved):
LGTM, tested locally and the sorts on both the Users table and the MUA roles table worked fine!

Hyperkid123 (approved):

PR #671, opened 2021-02-03
Add name validation for edit group

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-12127


john-dupuy (approved):
I was able to verify that the name validation is working. 

I noticed that the "Save" button is enabled even when a duplicate name is entered, and only disables once the user clicks the button. In the create group/role wizard, the "Next" button only enables after the validation is complete, and I think we should have that same behavior here.

I don't think it needs to be done in this PR however.

PR #674, opened 2021-02-09
Convert add group to ddf

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-12035


john-dupuy (approved):
LGTM, I ran some tests against this locally and they all passed. The only thing I had to update was that the `Confirm` button was changed to `Submit` at the end of the wizard.

Hyperkid123:
> LGTM, I ran some tests against this locally and they all passed. The only thing I had to update was that the `Confirm` button was changed to `Submit` at the end of the wizard.

That is actually how it's supposed to be according to accessibility rules.

Hyperkid123 (approved):

PR #677, opened 2021-02-15
Fix group name validation

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-12166

- fixed group's name validation so that submit button is enabled only if the form is valid
- removed unnecessary modal on loading and replaced with skeletons
- added ModalFormTemplate usage (+ made some further refactoring and code cleaning)
- fixed wrong Add group wizard's title 
- updated tests

@john-dupuy please verify not only the edit group modal but also group name validation in Add group wizard. The changes affect it as well so that we make sure that everything really works. Thank you

### DIFF
--- a/src/smart-components/group/validators.js
+++ b/src/smart-components/group/validators.js
@@ -15,7 +15,7 @@ const asyncValidator = async (groupName, idKey, id) => {
     return undefined;
   });
 
-  if (response?.data?.some((item) => item[idKey] !== id)) {
+  if (id ? response?.data?.some((item) => item[idKey] !== id) : response?.data?.length > 0) {
     throw 'Name has already been taken.';
   }
 
--- a/src/smart-components/group/validators.js
+++ b/src/smart-components/group/validators.js
@@ -1,7 +1,7 @@
 import { fetchGroups } from '../../helpers/group/group-helper';
 import asyncDebounce from '../../utilities/async-debounce';
 
-export const asyncValidator = async (groupName) => {
+const asyncValidator = async (groupName, idKey, id) => {
   if (!groupName) {
     return undefined;
   }
@@ -15,11 +15,11 @@ export const asyncValidator = async (groupName) => {
     return undefined;
   });
 
-  if (response?.data?.length > 0) {
-    throw 'Name has already been taken';
+  if (response?.data?.some((item) => item[idKey] !== id)) {
+    throw 'Name has already been taken.';
   }
 
   return undefined;
 };
 
-export const debouncedAsyncValidator = asyncDebounce(asyncValidator);
+export const debouncedAsyncValidator = asyncDebounce((value, idKey, id) => asyncValidator(value, idKey, id));
--- a/src/smart-components/group/add-group/add-group-wizard.js
+++ b/src/smart-components/group/add-group/add-group-wizard.js
@@ -12,6 +12,7 @@ import { addGroup } from '../../../redux/actions/group-actions';
 import SetRoles from './set-roles';
 import SetUsers from './set-users';
 import SummaryContent from './summary-content';
+import { debouncedAsyncValidator } from '../validators';
 
 const FormTemplate = (props) => <Pf4FormTemplate {...props} showFormControls={false} />;
 
@@ -78,6 +79,10 @@ const AddGroupWizard = ({ postMethod }) => {
       });
   };
 
+  const validatorMapper = {
+    'validate-group-name': ({ idKey, id }) => (value) => debouncedAsyncValidator(value, idKey, id),
+  };
+
   return cancelWarningVisible ? (
     <WarningModal
       type="group"
@@ -90,6 +95,7 @@ const AddGroupWizard = ({ postMethod }) => {
       schema={schema}
       subscription={{ values: true }}
       FormTemplate={FormTemplate}
+      validatorMapper={validatorMapper}
       componentMapper={{ ...componentMapper, ...mapperExtension }}
       onSubmit={onSubmit}
       initialValues={groupData}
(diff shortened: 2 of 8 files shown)


john-dupuy (changes requested):
I have two small small comments:
- User must click next/submit for the "Name taken" text to appear, it should appear automatically if the name is taken
- In the group wizard, if a user inputs a taken name, the next button until they clear the whole text box

src/smart-components/group/validators.js
```diff
@@ -1,25 +1,24 @@
...
+          throw 'Name has already been taken.';
+        }
+      })
+    : Promise.reject('Can have maximum of 150 characters.');
 
-  if (response?.data?.length > 0) {
-    throw 'Name has already been taken';
+export const uniqueNameValidator = asyncDebounce((value, idKey, id) => {
+  if (!value || value.length === 0) {
+    return Promise.reject('Required');
```
PanSpagetka, line 20: You are duplicating required check.

src/smart-components/group/validators.js
```diff
@@ -1,25 +1,24 @@
...
+    ? fetchGroups({ limit: 10, offset: 0, name, nameMatch: 'exact' }).then(({ data }) => {
+        if (data.length === 0) {
+          return undefined;
+        }
 
-  const response = await fetchGroups({ limit: 10, offset: 0, name: groupName, nameMatch: 'exact' }).catch((error) => {
-    console.error(error);
-    return undefined;
-  });
+        const taken = data.some((item) => item[idKey] !== id && item.name === name);
```
PanSpagetka, line 11: Here you are kinda duplicating logic that is done by API when you pass `nameMatch: "exact"`

PanSpagetka (commented):
Mostly code looks good, it fixes the issue. However I think that validator after changes is slightly more complex and I don't see any value in that.

src/smart-components/group/validators.js
fhlavac: yeah, thanks, siplified it

fhlavac:
@john-dupuy 
Issue 1 is really weird. However, in other similar modals (like "edit role") it does not show the error text after the initial load as well. If you click somewhere else and then try it again, it starts working as expected. I'll try asking DDF guys tomorrow.

Issue 2 - I was unable to reproduce it every time but it seems to be fixed now. Please try.

john-dupuy:
@fhlavac I am still able to reproduce issue 2. If you add a character to your duplicated name, the next button enables. But if you take away a character (or multiple) the next button does not enable.

john-dupuy (approved):
Thanks Filip, checking again today and issue 2 seems resolved! Let me know what the DDF people say about issue 1, I think we can just do it in a follow-up.

fhlavac:
@john-dupuy yeah, sometimes it works and sometimes not, but marking as WIP. I'm giving it one more try :smile:

Hyperkid123:
> Issue 1 is really weird. However, in other similar modals (like "edit role") it does not show the error text after the initial load as well. If you click somewhere else and then try it again, it starts working as expected. I'll try asking DDF guys tomorrow.

@john-dupuy errors are displayed only if the field was touched (after the field has lost focus), or after the submission attempt has failed. That is how the error display works. We could add "validateOnMount" functionality which has been deprecated when migrating from PF3. Basically we need a UX person to decide.

fhlavac:
@john-dupuy so issue 2 should be really fixed now, I left one little mistake in the previous version causing wrong validation when adding a group so this should be fixed as well. 

I was also asking about issue 1 and it should be behaving this way. It's how DDF validator is implemented. The user is expected to enter correctly on the first try so he should not be disturbed with error messages during the initial typing. But ...when the input loses focus and receives it again - user is editing the value - then he is receiving error messages as he is typing.

john-dupuy:
Thanks @fhlavac for the info! 

@Hyperkid123 that makes sense. In this context however, imo immediate feedback to the user is preferable, as this is how some text boxes work e.g. on password validation. I agree about getting UX feedback here, @mmenstr what do you think? 

Issue 1 is described as follows (referring to the edit group modal and the create group wizard)
> User must click next/submit for the "Name taken" text to appear, it should appear automatically if the name is taken

mmenestr:
Hmmm I want to say I agree with only showing an error on loss of focus, but at the same time I guess I can see a situation where a user doesn't necessarily click away from the input box until they try clicking a disabled next button..and THEN get the error which is kind of annoying.
I think my only worry about having errors show up while a user is typing is the potential annoyance - but I guess if they're only being shown an error and no green validation, then it probably wouldn't occur very often anyway?

Hyperkid123:
@mmenestr we can set it up to be field-specific. It does not have to apply to every field in the form.

mmenestr:
@Hyperkid123 Yeah I figured! How fast would the validation happen if they were to type in a username that is taken? I'm just wondering if they'll get errors while they're typing which would be annoying..

Hyperkid123:
The first error message would appear whenever the API call is finished. Cant give you an exact moment.

mmenestr:
Ok let's just do the validation while typing then, so they don't have to lose focus in order to know whether their role name is taken or not.

Hyperkid123:
@fhlavac we will merge this one and update the validation once the changes are made in DDF. It will be ready before this sprint ends.

fhlavac:
So I've updated edit-group modal so user is now gonna see a duplicate name error as he is typing. Unfortunately, the onMount validation cannot be enabled when adding a new group in a wizard since there would be "Required" error visible - the name input has no initial value.
PR: https://github.com/RedHatInsights/insights-rbac-ui/pull/693

PR #687, opened 2021-02-17
Update deps and config to use module in MUA

karelhala:
### Change styling usage

Since we changed the way how styles are being applied to a application we no longer need prefixed styling. Also we want to add `sassPrefix` to config, so all styles are applied to MUA as well.


karelhala:
@epwinchell I've removed the scope `.page__myUserAccess` from MUA home sass. If you think it was required so these styles won't effect rbac we could add prefix `.my-user-access`, but I'd rahter make it simple and looking at the styles none looks like they could be applied to rbac as well.

Hyperkid123 (approved):
Tested locally. Styling was fixed.

I have forgotten that we are not ignoring the proxy when adding deployment and there is missing the beta start script, but I will take care of it in another PR.

Hyperkid123:
@epwinchell I am gonna merge this so the ci is not looking funny. We can add additional styling changes later on.

PR #692, opened 2021-02-18
Serve from http

Hyperkid123:
Webpack proxy can route to https because of invalid certs


fhlavac (approved):

fhlavac:
Thank you! @Hyperkid123

PR #694, opened 2021-02-19
Validate group name on mount (add-group-wizard)

fhlavac:
Enabled onMount validation in add group wizard

### DIFF
--- a/src/smart-components/group/add-group/schema.js
+++ b/src/smart-components/group/add-group/schema.js
@@ -1,5 +1,9 @@
+import React from 'react';
 import validatorTypes from '@data-driven-forms/react-form-renderer/dist/esm/validator-types';
 import componentTypes from '@data-driven-forms/react-form-renderer/dist/esm/component-types';
+import ReviewTemplate from './review-template';
+import ReviewStepButtons from '../../common/review-step-buttons';
+import { AddGroupWizardContext } from './add-group-wizard';
 
 export default {
   fields: [
@@ -63,6 +67,9 @@ export default {
         {
           name: 'review',
           title: 'Review details',
+          // eslint-disable-next-line react/display-name
+          buttons: (props) => <ReviewStepButtons {...props} context={AddGroupWizardContext} />,
+          StepTemplate: ReviewTemplate,
           fields: [
             {
               component: 'summary-content',
--- a/src/smart-components/group/add-group/schema.js
+++ b/src/smart-components/group/add-group/schema.js
@@ -13,16 +13,13 @@ export default {
       fields: [
         {
           name: 'name-and-description',
-          title: 'Name and Description',
           nextStep: 'add-roles',
+          title: 'Name and Description',
           fields: [
             {
-              component: componentTypes.TEXT_FIELD,
+              component: 'set-name',
               name: 'group-name',
-              label: 'Group name',
-              isRequired: true,
               validate: [
-                { type: 'validate-group-name' },
                 {
                   type: validatorTypes.REQUIRED,
                 },
@@ -31,8 +28,7 @@ export default {
             {
               component: componentTypes.TEXTAREA,
               name: 'group-description',
-              type: 'text',
-              label: 'Group description',
+              hideField: true,
               validate: [
                 {
                   type: validatorTypes.MAX_LENGTH,
--- a/src/smart-components/group/add-group/set-name.js
+++ b/src/smart-components/group/add-group/set-name.js
@@ -1,0 +1,81 @@
+import React, { Fragment, useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Stack, StackItem } from '@patternfly/react-core';
+import useFieldApi from '@data-driven-forms/react-form-renderer/dist/esm/use-field-api';
+import useFormApi from '@data-driven-forms/react-form-renderer/dist/esm/use-form-api';
+import { TextInput } from '@patternfly/react-core/dist/esm/components/TextInput/TextInput';
+import { TextArea } from '@patternfly/react-core/dist/esm/components/TextArea/TextArea';
+import { FormGroup } from '@patternfly/react-core/dist/esm/components/Form/FormGroup';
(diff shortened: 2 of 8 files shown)


src/smart-components/group/add-group/set-name.js
```diff
@@ -0,0 +1,81 @@
...
+const SetName = (props) => {
+  const { input } = useFieldApi(props);
+  const formOptions = useFormApi();
+  const { 'group-name': name, 'group-description': description } = formOptions.getState().values;
+  const [groupName, setGroupName] = useState(name || '');
+  const [groupNameError, setGroupNameError] = useState();
+  const [groupDescription, setGroupDescription] = useState(description);
+
+  const groupNameValidated = () => (groupName === undefined || groupNameError ? 'error' : 'default');
+  const groupDescriptionValidated = () => (groupDescription && groupDescription.length > 150 ? 'error' : 'default');
```
karelhala, line 20: You can move these functions above component so they will not be re-created on each render.

karelhala, line 20: You can use safe accessor
```JS
groupDescription?.length > 150 ? 'error' : 'default'
```

src/smart-components/group/add-group/set-name.js
```diff
@@ -0,0 +1,81 @@
...
+        setGroupNameError(undefined);
+      })
+      .catch((error) => {
+        setGroupNameError(error);
+      });
+    setGroupName(value);
+  };
+
+  useEffect(() => {
+    groupName.length > 0 && processGroupName(groupName);
```
karelhala, line 36: Better to use safe accessor
```suggestion
    groupName?.length > 0 && processGroupName(groupName);
```

src/smart-components/group/add-group/set-name.js
```diff
@@ -0,0 +1,81 @@
...
+  }, []);
+
+  return (
+    <Fragment>
```
karelhala, line 40: No need to wrap in `Fragment`

src/smart-components/group/add-group/set-name.js
```diff
@@ -0,0 +1,81 @@
...
+      <Stack hasGutter>
+        <StackItem className="ins-c-rbac__summary">
+          <FormGroup label="Group name" helperTextInvalid={groupName ? groupNameError : 'Required'} isRequired validated={groupNameValidated()}>
+            <TextInput
+              value={groupName}
+              type="text"
+              validated={groupNameValidated()}
+              onBlur={() => {
+                groupName === '' && setGroupName(undefined);
+              }}
```
karelhala, line 50: No need for function body
```suggestion
              onBlur={() => groupName === '' && setGroupName(undefined)}
```

src/smart-components/group/add-group/set-name.js
```diff
@@ -0,0 +1,81 @@
...
+              onChange={(value) => processGroupName(value)}
+              aria-label="Group name"
+            />
+          </FormGroup>
+        </StackItem>
+        <StackItem>
+          <FormGroup label="Group description" helperTextInvalid="Can have maximum of 150 characters." validated={groupDescriptionValidated()}>
```
karelhala, line 57: @mmenestr is the text correct? `Can have maximum of 150 characters.` I think it would be better to use `Group description can have maximum of 150 characters`

src/smart-components/group/add-group/set-name.js
```diff
@@ -0,0 +1,81 @@
...
+
+  useEffect(() => {
+    groupName.length > 0 && processGroupName(groupName);
+  }, []);
+
+  return (
+    <Fragment>
+      <Stack hasGutter>
+        <StackItem className="ins-c-rbac__summary">
+          <FormGroup label="Group name" helperTextInvalid={groupName ? groupNameError : 'Required'} isRequired validated={groupNameValidated()}>
```
karelhala, line 43: The fallback should be `Group name is required`

fhlavac:
@karelhala @mmenestr All the error texts are as they were before, I did not change them in this PR. If there is a need for change I will change it also in the edit group modal to be in sync.

src/smart-components/group/add-group/schema.js
```diff
@@ -31,8 +28,7 @@ export default {
             {
               component: componentTypes.TEXTAREA,
               name: 'group-description',
-              type: 'text',
-              label: 'Group description',
+              hideField: true,
```
Hyperkid123, line 31: Is this supposed to be hidden?

fhlavac: @Hyperkid123 yeah, it is. The field is there because of validation for wizard buttons. Visible input is in the step custom component. I am not sure if this is the best solution but it is the only one that worked.

fhlavac:
Also added missing review step error, which was accidentally removed in https://github.com/RedHatInsights/insights-rbac-ui/pull/674


@PanSpagetka @john-dupuy Please check if it is correct now.

john-dupuy (approved):
LGTM, tested locally and it's working - thanks Filip!

src/smart-components/group/add-group/schema.js
```diff
@@ -67,6 +67,9 @@ export default {
         {
           name: 'review',
           title: 'Review details',
+          // eslint-disable-next-line react/display-name
+          buttons: (props) => <ReviewStepButtons {...props} context={AddGroupWizardContext} />,
```
Hyperkid123, line 71: Is there any specific reason why you can't use `useContext` inside the `AddGroupWizardContext` component?

fhlavac: @Hyperkid123 The ReviewStepButtons component is the same for the add-group wizard as for the add-role wizard. So in the ReviewStepButtons I need to pass a correct context to the useContext hook. Is there any other way to achieve the same thing? I may be using it wrong.

Hyperkid123: Ok. We can leave it for now. We can think of some nice way of switching the context later.

Hyperkid123 (approved):

PR #706, opened 2021-03-09
Add empty states on invalid identifier in URL

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-11538

added empty states showing on invalid identifier in URL to:
- groups
- group roles
- group members
- roles
- users

image01
image02
image03

@mmenestr


mmenestr:
Empty state texts look good, only thing is that it shouldn't be inside the page header, it should be in the page body! 
So the empty state should be on a grey background, while the breadcrumbs should be in the white page header!

fhlavac:
No problem, I'll change it :-)

john-dupuy (approved):
Tested out locally and functionally this works great, I will add some test cases for these in the near future. 

Thanks @fhlavac!

src/test/smart-components/group/principal/__snapshots__/principals.test.js.snap
```diff
@@ -194,18 +194,7 @@ Array [
...
-    Array [
-      "Add a user to configure user access.",
-      "",
-    ]
-  }
-  icon={[Function]}
-  title="There are no members in this group"
-/>
-`;
+exports[`<GroupPrincipals /> should render correctly with empty members 1`] = `null`;
```
karelhala, line 197: Is this change correct? I can't see changes to `GroupPrincipals` and this snapshot has changed.

fhlavac: No, it isn't, good catch, thank you. There was an outdated EmptyState component name in the expect statement.

fhlavac:
@mmenestr no problem, I've updated the screenshots. What do you think about titles in the header? I think it's better than leaving only breadcrumbs there.

fhlavac:
@mmenestr ok, it should be fixed. Updated the snapshots.

src/smart-components/group/group.js
```diff
@@ -25,21 +26,26 @@ const Group = ({
...
+  const history = useHistory();
+
+  const breadcrumbsList = () => [
+    { title: 'Groups', to: '/groups' },
+    groupExists ? { title: isFetching ? undefined : group.name, isActive: true } : { title: 'Invalid group', isActive: true },
+  ];
 
   const fetchData = (apiProps) => {
-    fetchGroup(apiProps);
+    fetchGroup(apiProps).then(({ value }) => {
```
PanSpagetka, line 46: I think it would be better to handle this in the reducer. Same for roles and users.

PanSpagetka (changes requested):

mmenestr:
Looks perfect thank you!!

src/smart-components/group/group.js
fhlavac: Moved the error flag to reducers, much better. Thanks :-)

PanSpagetka (approved):

PR #712, opened 2021-03-10
Replace addRole object with element

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-12850

replaced "add role" button object with jsx element


john-dupuy (approved):
LGTM, tested out locally and this seems to fix the bug.

Thanks Filip!

Hyperkid123 (approved):

PR #719, opened 2021-03-16
Fix empty state, reload on submit

PanSpagetka:
We can use default empty state of `TableToolbarView` and last step should now correctly redirect on role detail page and reload data.


fhlavac (approved):
Looks good to me!

src/smart-components/role/add-role-permissions/add-role-permission-success.js
```diff
@@ -14,13 +17,15 @@ const AddRolePermissionSuccess = ({currentRoleID}) => {
...
         </Link>
       </EmptyState>
     </>
   );
 };
 
 export default AddRolePermissionSuccess;
+
+AddRolePermissionSuccess.propTypes = {
+  currentRoleID: PropTypes.string,
```
Hyperkid123, line 30: I think this should be required.

src/smart-components/role/role-permissions.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
Hyperkid123, line 1: Can you remove this and fix the lint errors please?

PR #723, opened 2021-03-19
Convert add permissions to DDF, reuse steps from Create custom roles

PanSpagetka:
Convert add permissions to DDF, reuse steps from Create custom roles


src/smart-components/role/add-role-permissions/add-role-permissions-summary-content.js
```diff
@@ -38,21 +45,34 @@ const AddRolePermissionSummaryContent = ({ selectedPermissions = [], role = {} }
...
-            <Text component={TextVariants.p}>
+            <TextContent component={TextVariants.p}>
               {selectedPermissions.map((permission, index) => (
                 <Text key={index}> {permission.uuid} </Text>
               ))}
-            </Text>
+            </TextContent>
           </GridItem>
+          {hasCostResources && (
+            <>
```
karelhala, line 55: Just a nit pick, but can you use `Fragment` instead of `<>` I tend to overlook this.

src/smart-components/role/add-role-permissions/add-role-permissions-summary-content.js
```diff
@@ -38,21 +45,34 @@ const AddRolePermissionSummaryContent = ({ selectedPermissions = [], role = {} }
...
+              <GridItem span={3}>
+                <Text className="pf-c-title" component={TextVariants.h6}>
+                  Resource definitions:
+                </Text>
+              </GridItem>
+              <GridItem span={9}>
+                <TextContent component={TextVariants.p}>
+                  {resourceDefinitions
+                    .map((def) => def.resources)
+                    .map((resource, index) => (
```
karelhala, line 65: I don't think you need to have two maps.
```suggestion
                    .map(({ resources }, index) => (
```

Unless the `resources` are array, and if that's true then this won't help either. And you should import `flatten` from lodash.

PanSpagetka: Yep, there are two arrays:
```
[{
  permission: 'str1'
  resources: ['res1', res2', ...]
}, {,...
}]
```

karelhala: What exactly is input and what is the expected output?

src/smart-components/role/add-role-permissions/add-role-permissions-summary-content.js
```diff
@@ -38,21 +46,32 @@ const AddRolePermissionSummaryContent = ({ selectedPermissions = [], role = {} }
...
+          {hasCostResources && (
+            <React.Fragment>
+              <GridItem span={3}>
+                <Text className="pf-c-title" component={TextVariants.h6}>
+                  Resource definitions:
+                </Text>
+              </GridItem>
+              <GridItem span={9}>
+                <TextContent component={TextVariants.p}>
+                  {flatten(resourceDefinitions.map((def) => def.resources)).map((resource, index) => (
```
karelhala, line 64: I don't think this will work either
```JSX
{
  resourceDefinitions.map(
    ({ resources }) => resources.map((resource, index) => <Text key={index}> {resource} </Text>)
  )
}
```

PanSpagetka: It is working, but your suggestion is nicer

karelhala: The reason why it's working is because you probably tested it on single item resources. If there would be more than 1, they'd all be in single Text component.

karelhala (approved):

PR #729, opened 2021-03-24
Preserve pagination limit on adding group/role

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-12885

Fixed preserving pagination limit on adding, removing, and editing group/role

@john-dupuy


PanSpagetka (changes requested):
I tested this locally and it fixes the pagination issue, but you can't cancel Create Role wizard. Clicking on cancel just show notification, but wizard stays open.

fhlavac:
Pushed the fixed import, thank you @PanSpagetka

PanSpagetka (approved):

PR #748, opened 2021-04-09
Persist filters and pagination in Users

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-13427
https://issues.redhat.com/browse/RHCLOUD-13426

- Created functions for pagination and filters sync with the URL
- Applied new filtering and pagination persisting to RBAC Users
- Updated tests

As long as a user remains within the "User Access" application, filters persist, regardless of whether they are editing/deleting a group/role. Pagination and filters in Group members and Add group members remain not persistent and params are not visible in URL. @mmenestr please, let me know if this behavior is right.

pagination

@john-dupuy

### DIFF
--- a/src/helpers/shared/filters.js
+++ b/src/helpers/shared/filters.js
@@ -1,15 +1,13 @@
-export const getFiltersFromUrl = (keys, defaultValues = {}) => {
-  const searchParams = new URLSearchParams(location.search);
+export const getFiltersFromUrl = (history, keys, defaultValues = {}) => {
+  const searchParams = new URLSearchParams(history.location.search);
 
-  let filters = {};
-
-  keys.forEach((key) => {
+  let filters = keys.reduce((acc, key) => {
     const values = searchParams.getAll(key);
-    filters = {
-      ...filters,
+    return {
+      ...acc,
       [key]: values.length > 1 ? values : values[0],
     };
-  });
+  }, {});
 
   Object.keys(defaultValues).forEach((key) => {
     const value = defaultValues[key];
@@ -27,12 +25,12 @@ export const getFiltersFromUrl = (keys, defaultValues = {}) => {
     }
   });
 
-  window.history.replaceState({}, '', `${location.pathname}?${searchParams.toString()}`);
+  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
   return filters;
 };
 
-export const setFiltersToUrl = (newValues) => {
-  const searchParams = new URLSearchParams(location.search);
+export const setFiltersToUrl = (history, newValues) => {
+  const searchParams = new URLSearchParams(history.location.search);
   Object.keys(newValues).forEach((key) => searchParams.delete(key));
 
   Object.keys(newValues).forEach((key) => {
@@ -45,5 +43,5 @@ export const setFiltersToUrl = (newValues) => {
     }
   });
 
-  window.history.replaceState({}, '', `${location.pathname}?${searchParams.toString()}`);
+  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
 };
--- a/src/helpers/shared/filters.js
+++ b/src/helpers/shared/filters.js
@@ -1,4 +1,4 @@
-export const getFiltersFromUrl = (history, keys, defaultValues = {}) => {
+export const syncDefaultFiltersWithUrl = (history, keys, defaults = {}) => {
   const searchParams = new URLSearchParams(history.location.search);
 
   let filters = keys.reduce((acc, key) => {
@@ -9,8 +9,8 @@ export const getFiltersFromUrl = (history, keys, defaultValues = {}) => {
     };
   }, {});
 
-  Object.keys(defaultValues).forEach((key) => {
-    const value = defaultValues[key];
+  Object.keys(defaults).forEach((key) => {
+    const value = defaults[key];
     filters = {
       ...filters,
       [key]: Array.isArray(filters[key])
@@ -25,11 +25,14 @@ export const getFiltersFromUrl = (history, keys, defaultValues = {}) => {
     }
   });
 
-  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
+  history.replace({
+    pathname: history.location.pathname,
+    search: searchParams.toString(),
+  });
   return filters;
 };
 
-export const setFiltersToUrl = (history, newValues) => {
+export const applyFiltersToUrl = (history, newValues) => {
   const searchParams = new URLSearchParams(history.location.search);
   Object.keys(newValues).forEach((key) => searchParams.delete(key));
 
@@ -43,5 +46,8 @@ export const setFiltersToUrl = (history, newValues) => {
     }
   });
 
-  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
+  history.replace({
+    pathname: history.location.pathname,
+    search: searchParams.toString(),
+  });
 };
--- a/src/helpers/shared/filters.js
+++ b/src/helpers/shared/filters.js
@@ -1,0 +1,49 @@
+export const getFiltersFromUrl = (keys, defaultValues = {}) => {
+  const searchParams = new URLSearchParams(location.search);
+
+  let filters = {};
+
+  keys.forEach((key) => {
+    const values = searchParams.getAll(key);
+    filters = {
+      ...filters,
+      [key]: values.length > 1 ? values : values[0],
+    };
+  });
+
+  Object.keys(defaultValues).forEach((key) => {
+    const value = defaultValues[key];
+    filters = {
+      ...filters,
+      [key]: Array.isArray(filters[key])
+        ? [...new Set([...filters[key], ...(Array.isArray(value) ? value : [value])])]
+        : (value?.length > 0 && value) || filters[key],
+    };
+
+    if (Array.isArray(value)) {
+      value.forEach((item) => searchParams.getAll(key).includes(item) || searchParams.append(key, item));
+    } else {
+      searchParams.get(key) || (value && searchParams.set(key, value));
+    }
+  });
+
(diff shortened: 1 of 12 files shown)


src/helpers/shared/filters.js
```diff
@@ -0,0 +1,49 @@
+export const getFiltersFromUrl = (keys, defaultValues = {}) => {
+  const searchParams = new URLSearchParams(location.search);
+
+  let filters = {};
+
+  keys.forEach((key) => {
```
Hyperkid123, line 6: You can use `map` to directly initialize the filters variable.

src/helpers/shared/filters.js
```diff
@@ -0,0 +1,49 @@
...
+    const values = searchParams.getAll(key);
+    filters = {
+      ...filters,
+      [key]: values.length > 1 ? values : values[0],
```
Hyperkid123, line 10: Is it 100% certain that the `values[0]` entry will exist? `searchParams.getAll` can return an empty array.

src/helpers/shared/filters.js
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    if (Array.isArray(value)) {
+      value.forEach((item) => searchParams.getAll(key).includes(item) || searchParams.append(key, item));
+    } else {
+      searchParams.get(key) || (value && searchParams.set(key, value));
+    }
+  });
+
+  window.history.replaceState({}, '', `${location.pathname}?${searchParams.toString()}`);
```
Hyperkid123, line 30: It would be better if we use the `react-router-dom` history instance. We can have a globally accessible history instance. Checkout these docs: https://reactrouter.com/web/api/Router/history-object

src/helpers/shared/filters.js
```diff
@@ -0,0 +1,49 @@
...
+    const value = newValues[key];
+
+    if (Array.isArray(value)) {
+      value.forEach((item) => item && searchParams.append(key, item));
+    } else {
+      value && searchParams.set(key, value);
+    }
+  });
+
+  window.history.replaceState({}, '', `${location.pathname}?${searchParams.toString()}`);
```
Hyperkid123, line 48: ^^

src/helpers/shared/pagination.js
```diff
@@ -13,3 +13,22 @@ export const defaultCompactSettings = {
...
+  isNaN(parseInt(searchParams.get('page'))) && searchParams.set('page', initialLoad ? 1 : defaultPagination.offset / limit + 1);
+  const offset = (parseInt(searchParams.get('page')) - 1) * limit;
+
+  window.history.replaceState({}, '', `${location.pathname}?${searchParams.toString()}`);
+  return { limit, offset };
+};
+
+export const setPaginationToUrl = (limit, offset = 0) => {
+  const searchParams = new URLSearchParams(location.search);
+  searchParams.set('per_page', limit);
```
john-dupuy, line 31: @mmenestr Is there a preferred name for these pagination variables? Throughout the platform I see `offset` paired with `limit` (patch and systems table), `page` paired with `page_size` (CVE table). 

I don't think I've seen `per_page` anywhere.

fhlavac: @john-dupuy thanks for catching that, I think that using `page` number in the URL makes better sense to the user than `offset` value, I didn't realise that the limit value has different names throughout the platform so I'll rename it with any preferred name @mmenestr

mmenestr:
This sounds right to me!

src/helpers/shared/filters.js
fhlavac: No, it does not have to exist. But When the `getAll` function returns an empty array, then `values[0]` returns `undefined` which is not a problem in this case.

src/helpers/shared/filters.js
fhlavac: @Hyperkid123 please, could you explain more? `array.map` returns an array, but how to make it set object entries? I can use `reduce` but I am not sure how to do this with `map` efficiently.

fhlavac:
@Hyperkid123 all issues should be resolved

src/helpers/shared/filters.js
```diff
@@ -0,0 +1,47 @@
...
+    };
+
+    if (Array.isArray(value)) {
+      value.forEach((item) => searchParams.getAll(key).includes(item) || searchParams.append(key, item));
+    } else {
+      searchParams.get(key) || (value && searchParams.set(key, value));
+    }
+  });
+
+  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
```
Hyperkid123, line 28: Maybe I've misunderstood the purpose of this function, but why are we mutating the history search params in a function called `getFiltersFromUrl`? Shouldn't we change the search params in the other one?

Also, youou can use replace object as argument instead of constructing the whole query. Like so:
```jsx
history.replace({
  pathname: history.location.pathname,
  search: searchParams.toString()
})
```

src/helpers/shared/filters.js
```diff
@@ -0,0 +1,47 @@
...
+    const value = newValues[key];
+
+    if (Array.isArray(value)) {
+      value.forEach((item) => item && searchParams.append(key, item));
+    } else {
+      value && searchParams.set(key, value);
+    }
+  });
+
+  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
```
Hyperkid123, line 46: ^^
```jsx
history.replace({pathname, search})
```

src/helpers/shared/pagination.js
```diff
@@ -13,3 +13,22 @@ export const defaultCompactSettings = {
...
+
+export const getPaginationFromUrl = (history, defaultPagination, initialLoad) => {
+  const searchParams = new URLSearchParams(history.location.search);
+
+  isNaN(parseInt(searchParams.get('per_page'))) && searchParams.set('per_page', defaultPagination.limit);
+  const limit = parseInt(searchParams.get('per_page'));
+  isNaN(parseInt(searchParams.get('page'))) && searchParams.set('page', initialLoad ? 1 : defaultPagination.offset / limit + 1);
+  const offset = (parseInt(searchParams.get('page')) - 1) * limit;
+
+  history.replace(`${history.location.pathname}?${searchParams.toString()}`, {});
```
Hyperkid123, line 25: again, we are doing 'set' in a `get` function.

src/smart-components/group/add-group/users-list.js
```diff
@@ -63,13 +64,31 @@ const createRows = (userLinks) => (data, _expanded, checkedRows = []) => {
...
+
+  const [filters, setFilters] = useState({
+    username: (!inModal && stateFilters.username) || '',
+    email: (!inModal && stateFilters.email) || '',
+    status: !inModal && status ? status : ['Active'],
+  });
+
+  useEffect(() => {
+    inModal || setFilters(stateFilters);
+  }, [stateFilters]);
```
Hyperkid123, line 85: If `stateFilters` is a nested object, this trigger won't work properly and the effect will be run all the time. This structure will be always evaluated as changed: `{status: ['A']} !== {status: ['A']}`. We will be trigerring the effect even if it's not required. It is safer to break it down to literal values like so:

```jsx
  useEffect(() => {
    inModal || setFilters(stateFilters);
  }, [stateFilters.username, stateFilters.email, stateFilters.status]);
```

src/smart-components/group/add-group/users-list.js
```diff
@@ -78,6 +97,11 @@ const UsersList = ({ users, fetchUsers, isLoading, pagination, selectedUsers, se
     });
   };
 
+  const updateFilters = (payload) => {
+    inModal || updateUsersFilters(payload);
+    setFilters(payload);
```
Hyperkid123, line 102: Question. Do we keep the same state in two places? Can we use only one state storage? (it's fine if not) We wouldn't have to synchronize it.

src/smart-components/group/principal/add-group-members.js
```diff
@@ -69,7 +69,7 @@ const AddGroupMembers = ({
           </TextContent>
         </StackItem>
         <StackItem>
-          <CompactUsersList selectedUsers={selectedUsers} setSelectedUsers={setSelectedUsers} />
+          <CompactUsersList selectedUsers={selectedUsers} setSelectedUsers={setSelectedUsers} inModal={true} />
```
Hyperkid123, line 72: Unecessary constant inModal `true` value. This has the same affect
```jsx
<CompactUsersList selectedUsers={selectedUsers} setSelectedUsers={setSelectedUsers} inModal />
```

Hyperkid123 (commented):
The code looks functional. I have a couple of questions about its structure :slightly_smiling_face:

src/helpers/shared/filters.js
fhlavac: @Hyperkid123 the purpose is to get filter values from the URL and merge them with the default values provided. So maybe renaming the function to eg. `syncDefaultFiltersWithUrl` would be enough. I think that splitting the function does not really make sense there.

src/smart-components/group/add-group/users-list.js
fhlavac: oh, that makes sense, thank you, I'll fix that

src/smart-components/group/add-group/users-list.js
fhlavac: Removed usage of `useState`, it's actually not needed. This function is called to update filters in redux before data fetching actually succeeds. Otherwise, filter tags don't disappear immediately after clearing the filters but just after fetching the data.

Hyperkid123: If we want to have an immediate visual response, we can set the data within the `_PENDING` action and roll them back inf the request fails in the `_REJECTED` action. We can do it in a different PR.

Hyperkid123 (approved):
Tested in UI and it works like a charm. Nice job.

PR #753, opened 2021-04-14
Persist filters and pagination in Roles

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-13088

- simplified generic functions for pagination
- added pagination & filters persisting to Roles
- updated tests

filters-roles


src/smart-components/role/add-role-new/add-role-wizard.js
```diff
@@ -61,6 +63,8 @@ const AddRoleWizard = ({ pagination }) => {
     container.current.hidden = cancelWarningVisible;
   }, [cancelWarningVisible]);
 
+  const onClose = () => history.push(`${paths.roles}${createQueryParams({ page: 1, per_page: pagination.limit })}`);
```
Hyperkid123, line 66: If it won't refactoring other parts of the code already present in this PR, can we use the object argument instead of constructing the whole URL string?

```jsx
history.push({pathname, search, hash})
```

src/helpers/shared/helpers.js
```diff
@@ -40,3 +37,26 @@ export const selectedRows = (newSelection, isSelected) => (selected) => {
...
+    } else {
+      value && searchParams.set(key, value);
+    }
+  });
+  return `?${searchParams.toString()}`;
+};
+
+export const getBackRoute = (pathname, pagination, filters) => ({
+  pathname,
+  search: createQueryParams({
```
Hyperkid123, line 57: The `createQueryParams` returns a string with `?` prefix. That will cause the router url will contain `??` query prefix. The function should not return it.

src/smart-components/role/roles.js
```diff
@@ -29,40 +31,58 @@ const columns = [
...
-  const [filterValue, setFilterValue] = useState('');
   const dispatch = useDispatch();
   const { push } = useHistory();
-  const { roles, isLoading, pagination, userIdentity } = useSelector(selector, shallowEqual);
-  const fetchData = (options) => dispatch(fetchRolesWithPolicies(options));
+  const { roles, isLoading, filters, meta, userIdentity } = useSelector(selector, shallowEqual);
+  const fetchData = (options) => dispatch(fetchRolesWithPolicies({ ...options, inModal: false }));
+  const history = useHistory();
+
+  let pagination = roles.pagination || meta;
```
Hyperkid123, line 47: We should not be using global variable and then accessing/overriding it in an effect. You risk getting your data out of sync because the variable is overridden on each render. Use `useState` or `useRef` instead.

fhlavac: Thank you, moved to useEffect

Hyperkid123 (approved):

src/helpers/shared/helpers.js
fhlavac: @Hyperkid123 surprisingly, there was just one `?` - it was working as well. The react-router is probably handling it somehow but did not find exactly where. It's definitely better to delete it, thank you.

PR #754, opened 2021-04-14
feat(rbac): add access-requests

redallen:
Add `@redhat-cloud-services/access-requests-frontend` which has list and details pages and add them to RBAC. Filtering  and sorting isn't working on the list page yet, see https://github.com/RedHatInsights/insights-rbac/pull/465.

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -1,4 +1,5 @@
 /* global */
+const webpack = require('webpack');
 const { resolve } = require('path');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 const config = require('@redhat-cloud-services/frontend-components-config');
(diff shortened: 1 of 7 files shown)


Hyperkid123:
I've checked the UI at it works well. The only thing I would change would be the table loading state. We usually use skeletons instead of the spinner. But I believe some UX review will happen once this is merged :slightly_smiling_face:

redallen:
It's easy to switch to skeletons for the list and roles tables if needbe. cc @mmenestr who suggested the toolbar be hidden while loading.

Hyperkid123:
Yup, there is actually a FEC component for table loading state
```jsx
<SkeletonTable colSize={3} rowSize={10} />
```

redallen:
I've added skeletons instead of spinners for the loading states. This is now ready for review!

karelhala (approved):
We might want to switch to FED mods in access-request. We are going to share big chunks of UI which doesn't really need to be shared on NPM.

PR #783, opened 2021-04-30
Use FED moduled access request

karelhala:
### Register access request FED modules and use it

Since we can now share modules trough FED modules let's use them in access-requests app https://github.com/RedHatInsights/access-requests-frontend/pull/2, this requires registering such module before first use.

Also since access-request is using redux we have to pass registry to it so it can wrap each route with `Provider`.

### DIFF
--- a/src/smart-components/accessRequests/accessRequests.js
+++ b/src/smart-components/accessRequests/accessRequests.js
@@ -40,7 +40,6 @@ const AccessRequests = () => {
             scope="accessRequests"
             fallback={fallback}
             isInternal={false}
-            baseUrl={routes['access-requests']}
             getRegistry={getRegistry}
           />
         )}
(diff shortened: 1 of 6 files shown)


src/smart-components/accessRequests/accessRequests.js
```diff
@@ -1,24 +1,52 @@
...
+        path={routes['access-requests-detail']}
+        exact
+        render={() => (
+          <AsyncComponent
+            appName="access-requests"
+            module="./AccessRequestDetailsPage"
+            scope="accessRequests"
+            fallback={fallback}
+            isInternal={false}
+            baseUrl={routes['access-requests']}
```
redallen, line 43: You don't need this line anymore. I've fixed the URL to just be "../" in https://github.com/RedHatInsights/access-requests-frontend/pull/3

redallen (approved):
Since I've merged https://github.com/RedHatInsights/access-requests-frontend/pull/2 this can be merged.

Hyperkid123 (approved):

PR #786, opened 2021-05-10
Fix bulk select Add permission table

PanSpagetka:
When bulk select is partially selected, but can't select more, unselect all.

https://issues.redhat.com/browse/RHCLOUD-13983

### DIFF
--- a/src/smart-components/role/add-role-new/add-permissions.js
+++ b/src/smart-components/role/add-role-new/add-permissions.js
@@ -9,6 +9,7 @@ import { listPermissions, listPermissionOptions, expandSplats, resetExpandSplats
 import { getResourceDefinitions } from '../../../redux/actions/cost-management-actions';
 import { fetchRole } from '../../../redux/actions/role-actions';
 import { DisabledRowWrapper } from './DisabledRowWrapper';
+import { isEqual } from 'lodash';
 
 const columns = ['Application', 'Resource type', 'Operation'];
 const selector = ({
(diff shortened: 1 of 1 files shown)


src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -190,11 +190,10 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, isIn
...
 
   const setCheckedItems = (newSelection) => {
-    setSelectedPermissions(
-      newSelection(selectedPermissions)
-        .map(({ uuid }) => ({ uuid }))
-        .filter(({ uuid }) => getResourceType(uuid)?.count > 0)
-    );
+    const newSelected = newSelection(selectedPermissions)
+      .map(({ uuid }) => ({ uuid }))
+      .filter(({ uuid }) => uuid.split(':')[0] !== 'cost-management' || getResourceType(uuid)?.count > 0);
```
Hyperkid123, line 195: Surely there must be some other rule to identify non-interactive elements. How do we identify them in the table itself?

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -190,11 +190,12 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, isIn
...
-    setSelectedPermissions(
-      newSelection(selectedPermissions)
-        .map(({ uuid }) => ({ uuid }))
-        .filter(({ uuid }) => getResourceType(uuid)?.count > 0)
-    );
+    const newSelected = newSelection(selectedPermissions)
+      .filter(({ uuid, application }) => application !== 'cost-management' || getResourceType(uuid)?.count > 0)
+      .map(({ uuid }) => ({ uuid }));
+
+    const arrayEq = (a, b) => (a.length > b.length ? a.every(({ uuid }, i) => uuid === b[i]?.uuid) : b.every(({ uuid }, i) => uuid === a[i]?.uuid));
```
Hyperkid123, line 197: I think we could use `lodash.isEqual` instead.

PR #787, opened 2021-05-10
use custom template for Cost resource step

PanSpagetka:
it sets max-height for selects on Cost resource step. Ale styles for this step wasn't applied, so I fixed it by adding custom template. 

https://issues.redhat.com/browse/RHCLOUD-13983

### DIFF
--- a/src/smart-components/role/add-role-new/cost-resources-template.js
+++ b/src/smart-components/role/add-role-new/cost-resources-template.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Title } from '@patternfly/react-core';
+
+const CostResourcesTemplate = ({ formFields }) => (
+  <div className="rbac">
+    <Title headingLevel="h1" size="xl" className="pf-u-mb-lg">
+      Review details
+    </Title>
+    {formFields}
+  </div>
+);
+
+CostResourcesTemplate.propTypes = {
+  formFields: PropTypes.array,
+};
+
+export default CostResourcesTemplate;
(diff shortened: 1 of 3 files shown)


src/smart-components/role/add-role-new/cost-resources-template.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Title } from '@patternfly/react-core';
+
+const CostResourcesTemplate = ({ formFields }) => {
+  console.log('PRDEL', formFields);
```
Hyperkid123, line 6: :eye:

PanSpagetka: ups

src/smart-components/role/add-role-new/cost-resources.js
```diff
@@ -4,6 +4,7 @@ import { Select, SelectOption, SelectVariant, Grid, GridItem, Text, TextVariants
 import useFieldApi from '@data-driven-forms/react-form-renderer/dist/esm/use-field-api';
 import useFormApi from '@data-driven-forms/react-form-renderer/dist/esm/use-form-api';
 import { getResourceDefinitions, getResource } from '../../../redux/actions/cost-management-actions';
+import './add-role-wizard.scss';
```
Hyperkid123, line 7: Do we need this import? I would assume it was imported before. We should avoid duplicate imports.

src/smart-components/role/add-role-new/cost-resources-template.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Title } from '@patternfly/react-core';
+
+const CostResourcesTemplate = ({ formFields }) => (
+  <div className="rbac">
+    <Title headingLevel="h1" size="xl" className="pf-u-mb-lg">
+      Review details
+    </Title>
+    {[...formFields]}
```
Hyperkid123, line 10: Is it necessary to duplicate the array from props?

Hyperkid123 (approved):

PR #800, opened 2021-05-25
Fix group name validation & description change

fhlavac:
Fixes https://issues.redhat.com/browse/RHCLOUD-13081

- fixed group name validation to not block editing the description of a group while keeping the name the same
(validator was using `group.id`, which is undefined instead of `group.uuid` to make an exception for name collision with the group itself)
- also found and fixed a bug which disallowed to remove description from the group - when setting group description for example from "test description" to undefined, the description was not cleared

@john-dupuy

### DIFF
--- a/src/smart-components/group/edit-group-modal.js
+++ b/src/smart-components/group/edit-group-modal.js
@@ -50,7 +50,7 @@ const EditGroupModal = ({
   }, [group]);
 
   const onSubmit = (data) => {
-    const user_data = { ...data };
+    const user_data = { description: null, ...data };
     postMethod
       ? updateGroup(user_data)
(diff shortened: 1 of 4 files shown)


Hyperkid123 (commented):
The code looks good. We can wait for John's approval.

john-dupuy (commented):
Tested locally and verified the bug has been fixed -- thanks @fhlavac! One weird thing I noticed is that you have to click twice when putting in text to the "name" field on this modal. With "description" the second click is not needed.

Do you know what could be causing that?

fhlavac:
@john-dupuy it took a while but found it. Quite an interesting bug, the focus remained on the dropdown. Should be good now.

karelhala (approved):

PR #804, opened 2021-05-27
Migrate to npm 7

Hyperkid123:
### Changes
- enforce npm 15 and npm@7
- update all dependencies
- use webpack proxy for development
- fixed a bunch of invalid async tests that were creating actual API calls and crashing


karelhala (approved):

config/dev.webpack.config.js
```diff
@@ -6,9 +6,10 @@ const config = require('@redhat-cloud-services/frontend-components-config');
 const { config: webpackConfig, plugins } = config({
   rootFolder: resolve(__dirname, '../'),
   debug: true,
-  https: false,
   sassPrefix: '.rbac, .my-user-access',
-  ...(process.env.BETA && { deployment: 'beta/apps' }),
+  deployment: process.env.BETA ? 'beta/apps' : 'apps',
+  useProxy: true,
+  appUrl: process.env.BETA ? ['/beta/settings/my-user-access', '/beta/settings/rbac'] : ['/settings/my-user-access', '/settings/rbac'],
```
karelhala, line 12: Can you hide this behind `npm run start:proxy`? And document in README.md that we can run this either with webpack proxy or with insights proxy.

karelhala (changes requested):

config/dev.webpack.config.js
Hyperkid123: Is this really necessary? There is no benefit to using insights proxy. We can achieve exactly the same result with webpack proxy and we can add routes config to customize it. Which is exactly the same as configuring the spandx

karelhala: I still prefer using insights proxy (I have insights-proxy running on start with a lot of paths and just switch between app), at least have a task that runs this without webpack proxy.

Hyperkid123: OK

karelhala (approved):

PR #807, opened 2021-06-01
Fix group roles API call params order

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-14464


john-dupuy (approved):
tested locally and this fixes the issue, thanks @Hyperkid123!

PR #820, opened 2021-06-11
Fix "Group not found"error in Role detail

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-14401

The reason for this bug was an incomplete condition handling displaying the empty state.

@john-dupuy please verify with me, thanks for reporting!


john-dupuy (approved):
Thanks @fhlavac, I have verified the bug is fixed running this PR locally!

PR #823, opened 2021-06-16
Fetch the last page on non-existing page number in the URL

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-14050
When user enters a higher number of page to the URL than is the total number of pages, he does not see the default empty state but he is redirected to the really last page.
Works in Users, Group, Roles.

Also fixed not working pagination in the Group members.

https://user-images.githubusercontent.com/50696716/122757179-9fedf980-d297-11eb-8db3-200eaffc92ef.mp4

@john-dupuy 
@mmenestr

### DIFF
--- a/src/helpers/group/group-helper.js
+++ b/src/helpers/group/group-helper.js
@@ -20,17 +20,23 @@ export async function fetchGroups({
     groupApi.listGroups(limit, offset, filters.name, nameMatch, scope, username, uuid, roleNames, roleDiscriminator, orderBy, options),
     insights.chrome.auth.getUser(),
   ]);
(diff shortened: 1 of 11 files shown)


Hyperkid123 (approved):

john-dupuy:
@fhlavac in the video the per_page is set to 20, but it looks like only 5 roles are displayed when the user is redirected to the last page. Is this intended?

fhlavac:
@john-dupuy yeah, I would say yes. Per_page is set to 20 but 20 is not a divisor of 146. So the last page contains just 6 records.

john-dupuy:
Ah of course, that makes sense! Thanks @fhlavac

PR #870, opened 2021-07-29
Use display_name filter in roles

fhlavac:
This fix ensures using only `display_name` for filteing in roles


Hyperkid123 (approved):
Tested in UI. The name attribute is gone. :tada:

PR #891, opened 2021-08-24
OUIA for the add role button

john-dupuy:


src/smart-components/group/role/group-roles.js
```diff
@@ -157,7 +162,7 @@ const GroupRoles = ({
             to={`/groups/detail/${uuid}/roles/add_roles`}
             key="add-to-group"
           >
-            {addRoleButton(disableAddRoles)}
+            {addRoleButton(disableAddRoles, generateOuiaID(`${name}`))}
```
ryelo, line 165:
```suggestion
            {addRoleButton(disableAddRoles, generateOuiaID(name))}
```

ryelo (changes requested):

src/smart-components/group/role/group-roles.js
john-dupuy: When I do this I get an error ```Cannot read property 'toLowerCase' of undefined``` 

I used those backticks to match what was done to access e.g. the `uuid` of the group.

Hyperkid123: @john-dupuy with the backticks it would send a `"undefined"` string to the function (same as calling `undefined.toString()`) which is not something we probably don't want. So if the name is undefined, can we use some other value? Id or something?

john-dupuy: @Hyperkid123 The `name` is defined for every group (it's a required field https://github.com/RedHatInsights/insights-rbac/blob/2fd28158b8069eecf6ba13946133c68bd007eced/rbac/management/group/model.py#L41)

Hyperkid123: So what string is calling the `toLowerCase` method then?

john-dupuy: @Hyperkid123 in testing locally, `${name}` resolves to the actual name of the group (just like `${uuid}` does here:
https://github.com/RedHatInsights/insights-rbac-ui/blob/9fc923fb508f2517df8fea309f4a5c7ee513bbdf/src/smart-components/group/role/group-roles.js#L157)
So it'd be that string.

Hyperkid123: It might not be loaded on the first render, therefore the `undefined` error.  Does it work if you try `name || ''`. Just to verify.

john-dupuy: aha, thanks Martin, that does work.

Hyperkid123 (approved):

PR #919, opened 2021-09-13
Fix conditional filter styling in Role permissions

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-16088

Fixed broken conditional filter styling in Role permissions

### DIFF
--- a/src/smart-components/role/role-permissions.scss
+++ b/src/smart-components/role/role-permissions.scss
@@ -23,3 +23,12 @@
   }
   
 }
+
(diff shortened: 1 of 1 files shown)


karelhala (approved):

karelhala:
@fhlavac this might be a problem for other apps as well. We might want to port this one to FEC repo as well.

fhlavac:
@karelhala this is probably just local issue there, maybe because of some wrong import or so. Didn't find the actual reason yet 😕

PR #942, opened 2021-10-13
Dedupe list of permission when adding to a role

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-9209

When adding permissions to an existing role, permissions that are already in a role 
are excluded from being displayed in the table, not just disabled as before.

Before:

Now:

### DIFF
--- a/src/helpers/permission/permission-helper.js
+++ b/src/helpers/permission/permission-helper.js
@@ -2,8 +2,8 @@ import { getPermissionApi } from '../shared/user-login';
 
 const accessApi = getPermissionApi();
 
-export async function listPermissions(limit, offset, orderBy, application, resourceType, verb, permission, options) {
-  return await accessApi.listPermissions(limit, offset, orderBy, application, resourceType, verb, permission, options);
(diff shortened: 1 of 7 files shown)


src/helpers/permission/permission-helper.js
```diff
@@ -2,8 +2,8 @@ import { getPermissionApi } from '../shared/user-login';
 
 const accessApi = getPermissionApi();
 
-export async function listPermissions(limit, offset, orderBy, application, resourceType, verb, permission, options) {
-  return await accessApi.listPermissions(limit, offset, orderBy, application, resourceType, verb, permission, options);
+export async function listPermissions(limit, offset, orderBy, application, resourceType, verb, permission, excludeGlobals, excludeRoles, options) {
```
Hyperkid123, line 5: I see a change in the order of the arguments (options are now pushed back), but no client upgrade? Does this mean that the options object was never used or was the order incorrect all along?

fhlavac: Oh, I forgot to commit the client dependency update :facepalm:
Good catch, thank you @Hyperkid123

PR #974, opened 2021-11-15
Add a ErrorCatcher component.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16587

### TODO
- [x] fix tests


john-dupuy:
Testing this out locally, I see the following when I try to access the RBAC of the approver:
Screenshot from 2021-11-15 08-16-19

What should "undefined" be here? Furthermore, if you try to click "My User Access" from the left nav after hitting this page, it doesn't load and stays as "You don't have access to undefined". The user must refresh chrome in order for MUA to be displayed.

Hyperkid123:
> What should "undefined" be here?
@john-dupuy I think I forgot to add the name there. It should be RBAC.

> The user must refresh chrome in order for MUA to be displayed.
I will purge the state whenever the route changes

> Also, I was able to access the groups page, but I guess this is because the BE doesn't return a 403 here?
That is correct. The BE returns the data.

Hyperkid123:
@john-dupuy issues adressed.

src/helpers/shared/user-login.js
```diff
@@ -1,9 +1,36 @@
 /* eslint-disable no-unused-vars */
-import axiosInstance from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+import axios from 'axios';
+import {
+  authInterceptor,
+  responseDataInterceptor,
+  interceptor401,
+  interceptor500,
+  errorInterceptor,
+} from '@redhat-cloud-services/frontend-components-utilities/interceptors';
```
karelhala, line 9: Is there a reason why not to import the instance here and just pass in `interceptor403`? You are effectively doing the same thing, but on more lines.

src/presentational-components/shared/ErrorReducerCatcher.js
```diff
@@ -0,0 +1,39 @@
...
+const ErroReducerCatcher = ({ children }) => {
+  const errorCode = useSelector(({ errorReducer: { errorCode } }) => errorCode);
+  const { pathname } = useLocation();
+  const dispatch = useDispatch();
+  useEffect(() => {
+    if (errorCode) {
+      dispatch({ type: API_ERROR, payload: undefined });
+    }
+  }, [pathname]);
+  if (errorCode) {
```
karelhala, line 26: Empty line to make it look nicer.
```suggestion
  }, [pathname]);

  if (errorCode) {
```

src/helpers/shared/user-login.js
Hyperkid123: Yes. Because the order of interceptors matters. The common error interceptor changes the error payload. It is stripping some very important information - the status:
```js
throw { ...errObject.response.data, statusText: errObject.response.statusText };
```
Which makes it impossible to get the status code from the error payload. If we could change the order of interceptors, I would use the default instance.

karelhala: Ah I see, we might want to create `instanceBuilder` which would take an array of `request` interceptors whose will be used before using `authInterceptor` and an array of `response` interceptors whose will be used before using `interceptor401``

Hyperkid123: Well, the other interceptors don't really matter because they throw the original error. Only the last one is changing the payload.

john-dupuy (approved):

Hyperkid123:
@john-dupuy yes it is expected.

PR #985, opened 2021-11-23
Quickstarts doublepress

arivepr:
Fixed issue with 'toggle quickstart' needing a double press after closing the active quickstart drawer.
(not picture needed/applicable for this)


Hyperkid123 (commented):
@arivepr tested in UI and works fine. If you could remove just the extra commit and then we can merge.

Hyperkid123 (approved):

PR #989, opened 2021-12-08
Update warning modal close event.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-17120

@john-dupuy I was unable to reproduce this issue when running the code locally. But I've added changes around the warning modal close event because I have the following theory:

The cancel modal cancel action is pushed into the browser history. That history change forces the parent component (wizard modal) to close. That will destroy the parent focus trap of the cancel modal which will lead to the error we are seeing due to the nature of async rendering and React portals. However I can't confirm the theory because we can't reproduce the issue locally, only once the code was deployed.

I've added additional logic and pushed the history event to the end of the event loop, so the modal should properly close before the history change actually happens.

If this won't help, I will add a silent error boundary which will ignore errors with the `focus-trap` string and render nothing.

Unfortunately, we have to deploy the code to check if this change helped.


john-dupuy (approved):
Your theory makes sense to me - let's give it a try in stage-beta.

PR #990, opened 2021-12-08
Allow user access admins to edit/delete groups.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-17159

@john-dupuy we now allow users with  `rbac:*:*` permissions to see the actions.


john-dupuy (approved):
Tested locally and this works well - thanks Martin!

PR #994, opened 2021-12-09
Add silent error boundary to fix PF focus trap issues.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-17120

As we are unable to manually reproduce the issue and the automated tests are still failing, I am adding a silent error boundary that will ignore the `focus-trap` issue as the component is destroyed anyway when the problem is occurring. I will continue investigating the issue and try to come up with a propper fix.


john-dupuy (approved):
LGTM, let's give it a try.

PR #996, opened 2021-12-13
Fetch only allowed premissions in roles

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-16875

- Enabled fetching only allowed permissions in the "Add role" & "Add permission" wizards.
- Added "Tailored roles" - when a new role is being created as a copy of existing role with some permissions that cannot be copied, they won't be displayed and preselected in the "Add permission step" and user will be notified in that step via alert that says “The following permissions can not be added to a custom role and were removed from the copied role: XXX"

image

@mmenestr


src/smart-components/role/add-role-new/add-permissions-template.js
```diff
@@ -47,6 +49,16 @@ const AddPermissionTemplate = ({ formFields }) => {
           )}
         </Text>
       </TextContent>
+      {notAllowedBasePermissions?.length > 0 && !alertClosed ? (
```
karelhala, line 52: You should wrap these in parenthesis.
```suggestion
      {(notAllowedBasePermissions?.length > 0 && !alertClosed) ? (
```

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -182,7 +183,21 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...p
...
+    let notAllowed = [];
+
+    const basePermissions =
+      baseRole?.access.filter((item) => {
+        if (applicationOptions.includes(item?.permission?.split(':')[0])) {
+          return true;
+        } else {
+          notAllowed.push(item);
+        }
+        return false;
```
karelhala, line 195: No need for the `else` statement.
```suggestion
        }
        
        notAllowed.push(item);
        return false;
```

src/smart-components/role/add-role-new/add-permissions.js
```diff
@@ -182,7 +183,21 @@ const AddPermissionsTable = ({ selectedPermissions, setSelectedPermissions, ...p
...
+      }) || [];
+    formOptions.change(
+      'not-allowed-permissions',
+      notAllowed.map((item) => item.permission)
```
karelhala, line 199: You can use destruct instead of accessing the permission, it's efeectively the same, just a bit smoother to look at
```suggestion
      notAllowed.map(({ permission }) => permission)
```

karelhala (approved):
Looking good! Just a few nitpicks, but overall looks solid.

src/smart-components/role/add-role-new/add-permissions-template.js
fhlavac: @karelhala ESLint does not allow me to do it

Hyperkid123 (approved):

PR #1006, opened 2022-01-10
Call `toString` on user list API values.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-17403

The linked issue is not reproducible on only of our testing accounts. The error is most likely caused by some invalid DB value in the user's table, but we don't have access to the relevant data set. We can try calling the `toString` method on the values returned from API in the user's table.

This should help us and the user to find out if the data is corrupted without accessing their database.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,37 +1,16245 @@
 {
   "name": "insights-rbac-frontend",
   "version": "1.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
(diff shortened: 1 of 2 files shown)


karelhala (approved):
This will print `[object Object]` on places which will have more than just string. But I am fine with that, rather to print something than break the UI.

Hyperkid123:
@karelhala that is the point. There seem to be corrupted data which crashes the whole UI. And I would need access to the customer account in order to debug it.

PR #1013, opened 2022-01-13
Allow "User access admin" to add/remove group roles/members

fhlavac:
Fixes: https://issues.redhat.com/browse/RHCLOUD-17354

Fixed displaying add/remove group roles/members buttons to org admins and users with "User access admin" role using new `PermissionsContext` 

@john-dupuy

### DIFF
--- a/src/smart-components/group/principal/principals.js
+++ b/src/smart-components/group/principal/principals.js
@@ -1,6 +1,6 @@
 /* eslint-disable camelcase */
 import { nowrap } from '@patternfly/react-table';
-import React, { Fragment, useState, useEffect } from 'react';
+import React, { Fragment, useState, useEffect, useContext, useRef } from 'react';
 import { useDispatch, useSelector, shallowEqual } from 'react-redux';
 import { Link, Route, useHistory, useParams } from 'react-router-dom';
 import { TableToolbarView } from '../../../presentational-components/shared/table-toolbar-view';
@@ -12,6 +12,7 @@ import Section from '@redhat-cloud-services/frontend-components/Section';
 import RemoveModal from '../../../presentational-components/shared/RemoveModal';
 import UsersRow from '../../../presentational-components/shared/UsersRow';
 import paths from '../../../utilities/pathnames';
+import PermissionsContext from '../../../utilities/permissions-context';
 
 const columns = [
   { title: 'Status', transforms: [nowrap] },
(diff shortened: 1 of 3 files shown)


src/smart-components/group/principal/principals.js
```diff
@@ -74,7 +75,7 @@ const GroupPrincipals = () => {
   };
 
   const actionResolver = () =>
-    !(userIdentity && userIdentity.user && userIdentity.user.is_org_admin)
+    !(orgAdmin || userAccessAdministrator)
```
Hyperkid123, line 78: Can you consolidate the condition into a variable so we don't evaluate it all the time?

src/smart-components/group/role/group-roles.js
```diff
@@ -156,7 +157,7 @@ const GroupRoles = ({
   const history = useHistory();
 
   const toolbarButtons = () => [
-    ...(userIdentity && userIdentity.user && userIdentity.user.is_org_admin
+    ...(orgAdmin || userAccessAdministrator
```
Hyperkid123, line 160: ^^

Hyperkid123 (commented):
Almost perfect.

src/smart-components/group/role/group-roles.js
```diff
@@ -88,6 +88,8 @@ const GroupRoles = ({
   const [showRemoveModal, setShowRemoveModal] = useState(false);
   const [confirmDelete, setConfirmDelete] = useState(() => null);
   const [deleteInfo, setDeleteInfo] = useState({});
+  const { userAccessAdministrator, orgAdmin } = useContext(PermissionsContext);
+  const hasPermissions = useRef(orgAdmin || userAccessAdministrator);
```
karelhala, line 92: I like the usage of ref in here (no need to trigger yet another re-render when `orgAdmin` or `userAccessAdministrator` changes) I am a bit concerned about the change of these values. There is no effect looking at these calues and changing the current value of `hasPermissions`. I know these values can't change while sitting on the page, but what about initial render?

karelhala (commented):
Just a question about ref value update, but overall looks good

src/smart-components/group/role/group-roles.js
fhlavac: @karelhala good point, it could cause problems. I added `useEffect`, which monitors the value as you suggest. 
Thank you :slightly_smiling_face:

karelhala (approved):

PR #1023, opened 2022-01-31
Removed User Management text suggestion for non-org admins on Users page

radekkaluzik:
https://issues.redhat.com/browse/RHCLOUD-17454

### DIFF
--- a/src/presentational-components/shared/ActiveUsers.js
+++ b/src/presentational-components/shared/ActiveUsers.js
@@ -1,13 +1,19 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import { Text, TextVariants } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
-const ActiveUser = ({ description, linkTitle }) => {
+import PermissionsContext from '../../utilities/permissions-context';
+
+const ActiveUser = ({ linkDescription, linkTitle }) => {
   const env = insights.chrome.getEnvironment();
   const prefix = insights.chrome.isProd ? '' : `${env === 'ci' ? 'qa' : env}.`;
-  return (
+  const description = 'These are all of the users in your Red Hat organization. ';
+  const { orgAdmin } = useContext(PermissionsContext);
+
+  return orgAdmin ? (
     <Text className="pf-u-mt-0" component={TextVariants.h7}>
       {description}
+      {linkDescription}
       <Text
         component={TextVariants.a}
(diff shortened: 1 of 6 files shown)


aneelac22 (approved):
LGTM

aneelac22:
The build is failing though, unrelated errors

fhlavac:
@radekkaluzik as far I understand, the first part of the text ("These are all of the users in your Red Hat organization.") should be visible always and only "To add new users or manage existing users, go to your user management list" should be conditional, right? (cc @mmenestr)

fhlavac:
@radekkaluzik please, update the snapshots to fix the CI :slightly_smiling_face:

radekkaluzik:
Tests are fixed. Please don't merge yet. Need to clarify the question from Filip first.

fhlavac:
@Hyperkid123 not sure about the component API since we have to split the description into multiple parts. I would say now we duplicate the text too much. What do you think?

Hyperkid123:
@fhlavac I don't like splitting the text into multiple sections either but we don't have much choice in this I am afraid. We could pass it down as a `React.node` and use it. But then we just move the problem elsewhere.

fhlavac:
@radekkaluzik how about something like this? 

```
const ActiveUser = ({ linkDescription, linkTitle }) => {
  const env = insights.chrome.getEnvironment();
  const prefix = insights.chrome.isProd ? '' : `${env === 'ci' ? 'qa' : env}.`;
  const description = 'These are all of the users in your Red Hat organization. ';
  const { orgAdmin } = useContext(PermissionsContext);

  return orgAdmin ? (
    <Text className="pf-u-mt-0" component={TextVariants.h7}>
      {description}
      {linkDescription}
      <Text
        component={TextVariants.a}
        href={`https://www.${prefix}redhat.com/wapps/ugc/protected/usermgt/userList.html`}
        target="_blank"
        rel="noopener noreferrer"
      >
        {linkTitle}
        <ExternalLinkAltIcon />
      </Text>
      .
    </Text>
  ) : (
    <Text className="pf-u-mt-0" component={TextVariants.h7}>
      {description}
    </Text>
  );
};

ActiveUser.propTypes = {
  linkDescription: PropTypes.node,
  linkTitle: PropTypes.node,
};

ActiveUser.defaultProps = {
  linkDescription: '',
  linkTitle: ' user management list ',
};

export default ActiveUser;
```

radekkaluzik:
@fhlavac Reworked as you suggested. I have tested both cases.

non-org admin:
non-org admin

org admin:
org admin
.

PR #1036, opened 2022-02-10
Cleaning up 403 functionalaties

arivepr:
No longer rendering toast notifications for 403's, only showing the 403 empty state.


src/App.scss
```diff
@@ -8,6 +8,10 @@
     font-weight: var(--pf-global--FontWeight--bold);
 }
 
+.notifications-portal .toast-blocked {
```
Hyperkid123, line 11: What is this classame used for?

arivepr: nice catch! old code

Hyperkid123 (approved):

PR #1041, opened 2022-02-16
Adding condition to account for view permissions.

arivepr:
[Addressing this card.](https://issues.redhat.com/browse/RHCLOUD-17079)

Adding functionality to disable links for user-details whenever the active user does not have rbac admin permissions. 
Screen Shot 2022-02-16 at 5 36 03 PM
(ignore the quickstarts button up top)


src/smart-components/user/users.js
```diff
@@ -33,7 +36,7 @@ const Users = () => {
     <StackItem>
       <Section type="content" id={ 'users' }>
         <UsersList
-          userLinks
+          userLinks={(activeUserPermissions.userAccessAdministrator ? true : false)}
```
Hyperkid123, line 39: You can use the variable directory as it is always `true`/`false`.

Hyperkid123 (changes requested):

Hyperkid123 (approved):

PR #1049, opened 2022-02-25
Fix pagination redirect on invalid params

fhlavac:
fixes: https://issues.redhat.com/browse/RHCLOUD-18141

Added missing initial value to the `UrlSearchParams` constructor


aneelac22 (approved):

src/helpers/shared/pagination.js
```diff
@@ -15,7 +15,7 @@ export const calculatePage = (limit = defaultSettings.limit, offset = 0) => Math
 export const calculateOffset = (page = 1, limit = defaultSettings.limit) => (page - 1) * limit;
 
 export const syncDefaultPaginationWithUrl = (history, defaultPagination = defaultSettings) => {
-  let searchParams = new URLSearchParams();
+  let searchParams = new URLSearchParams(window.location.search);
```
Hyperkid123, line 18: Rather than the window location, can you use the history location from the parameter?

Hyperkid123 (approved):

PR #1073, opened 2022-03-21
Update default group routing

aneelac22:
[RHCLOUD-17856]


Hyperkid123:
Requires: https://github.com/RedHatInsights/javascript-clients/pull/150

aneelac22:
This PR is ready for review

Hyperkid123 (commented):
@aneelac22 the data fetching looks good. Can you also change the react route as described in the issue? There is an edge case where the default group route will not work because after you modify a default group the uuid changes.

PR #1104, opened 2022-04-13
Move deploy template to its own folder

BlakeHolifield:


bsquizz:
Other apps name it deploy/frontend.yaml do we want to stick w/ that convention?

ryelo (approved):

PR #1130, opened 2022-05-03
Update default group routing

aneelac22:
[RHCLOUD-17856]

### DIFF
--- a/src/smart-components/group/group-table-helpers.js
+++ b/src/smart-components/group/group-table-helpers.js
@@ -49,7 +49,7 @@ export const createRows = (isAdmin, data, _opened, selectedRows = []) =>
           <Fragment key={uuid}>
             <div className="pf-m-inline-flex">
               {isAdmin ? (
-                <Link key={`${uuid}-link`} to={`/groups/detail/${uuid}`}>
+                <Link key={`${uuid}-link`} state={{ uuid }} to={{ pathname: `/groups/detail/${isPlatformDefault ? 'default-access' : uuid}` }}>
                   {name}
                 </Link>
               ) : (
--- a/src/smart-components/group/group.js
+++ b/src/smart-components/group/group.js
@@ -7,7 +7,7 @@ import AppTabs from '../app-tabs/app-tabs';
 import { TopToolbar, TopToolbarTitle } from '../../presentational-components/shared/top-toolbar';
 import GroupPrincipals from './principal/principals';
 import GroupRoles from './role/group-roles';
-import { fetchGroup, fetchGroups } from '../../redux/actions/group-actions';
+import { fetchGroup, fetchGroups, fetchSystemGroup } from '../../redux/actions/group-actions';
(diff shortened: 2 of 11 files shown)


aneelac22:
doing some final clean up

src/smart-components/group/group-table-helpers.js
```diff
@@ -49,7 +49,11 @@ export const createRows = (isAdmin, data, _opened, selectedRows = []) =>
           <Fragment key={uuid}>
             <div className="pf-m-inline-flex">
               {isAdmin ? (
-                <Link key={`${uuid}-link`} to={`/groups/detail/${uuid}`}>
+                <Link
+                  key={`${uuid}-link`}
+                  state={{ uuid }}
+                  to={{ pathname: isPlatformDefault ? `/groups/detail/default-access` : `/groups/detail/${uuid}` }}
```
karelhala, line 55: I'd change the string in here so we don't repeat the URL
```suggestion
                  to={{ pathname: `/groups/detail/${isPlatformDefault ? 'default-access' : uuid}` }}
```

src/smart-components/group/group.js
```diff
@@ -28,6 +28,7 @@ const Group = ({
   isFetching,
   onDelete,
 }) => {
+  const isPlatformDefault = !(uuid !== 'default-access');
```
karelhala, line 31: No need to have two negative comparators
```suggestion
  const isPlatformDefault = uuid === 'default-access';
```

src/smart-components/group/group.js
```diff
@@ -63,10 +65,21 @@ const Group = ({
...
+      insights.chrome.appObjectId(uuid);
+      return () => insights.chrome.appObjectId(undefined);
+    } else {
+      if (systemGroupUuid) {
+        fetchData(systemGroupUuid);
+        insights.chrome.appObjectId(systemGroupUuid);
+        return () => insights.chrome.appObjectId(undefined);
+      } else {
+        fetchSystemGroup();
+      }
```
karelhala, line 80: I think that this could be simplified
```suggestion
    const currUuid = uuid !== 'default-access' ? uuid : systemGroupUuid;
    fetchData(currUuid);
    if (currUuid) {
      insights.chrome.appObjectId(currUuid);
      return () => insights.chrome.appObjectId(undefined);
    }
```

karelhala (commented):
Looking good, just a few cosmetic changes.

src/smart-components/group/group.js
aneelac22: I will make this change, but I guess it was unconscious choice I always try to check `!==` rather than `===` because the former is little more efficient, but I guess it doesn't make much difference here

Hyperkid123 (commented):
The tests are making real API requests. Yu have to mock the axios instance and fake the responses.

Hyperkid123 (approved):

PR #1139, opened 2022-05-10
[POC] WS data invalidation model

Hyperkid123:
requires: https://github.com/RedHatInsights/insights-chrome/pull/1895

### Description

This POC introduces data invalidation and caching using websockets. Data for RBAC groups is only fetched if the cache is non-existing or stale.

Please do not take the implementation seriously. We would need a proper client library to automate most of the setup. Also, the actual data store could be embedded into chrome. Something to think about. Read more about the WS in the linked chrome POC draft


karelhala:
Closing as stale ticket

PR #1156, opened 2022-06-03
Increase permissions fetch limit to 20000

fhlavac:
Fixes https://issues.redhat.com/browse/RHCLOUD-18338


Hyperkid123 (approved):

Hyperkid123:
I wonder if we can get some config like `-1` to just fetch all. Might be worth exploring that option.

PR #1162, opened 2022-06-07
Update document title

fhlavac:
Fixes https://issues.redhat.com/browse/RHCLOUD-19768

Updated document title for RBAC bundle and its MUA page


src/routes.js
```diff
@@ -12,19 +13,33 @@ const AccessRequests = lazy(() => import('./smart-components/accessRequests/acce
...
+      (Object.values(pathnames).find(
+        (item) =>
+          !!matchPath(location.pathname, {
+            path: item.path,
+            exact: true,
+            strict: false,
+          })
+      )?.title || 'User Access') + ' | console.redhat.com'
+    );
+  }, [location]);
```
Hyperkid123, line 28: You should only listen for the `location.pathname` changes. We don't wanna run the effect on each location change (search, hash, etc.)

PR #1195, opened 2022-07-25
Init translations

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-18504

- Initialized translations in RBAC - replaced strings with translations
- updated tests

image


src/locales/data.json
```diff
@@ -0,0 +1,262 @@
...
+    "removePermissionsQuestion": "Remoe permissions?",
+    "removePermissions": "Remove permissions",
+    "noRolePermissions": "There are no permissions in this role",
+    "definedResources": "Defined resources",
+    "userDescription": "{username}'s roles, groups and permissions.",
+    "userNotFound": "User not found",
+    "userNotFoundDescription": "User with username {username} does not exist.",
+    "addNewUsersText": "To add new users or manage existing users, go to your"
+  }
+}
```
aneelac22, line 262: I don't think this file is needed

src/AppEntry.js
```diff
@@ -4,19 +4,25 @@ import { Provider } from 'react-redux';
 import registry, { RegistryContext } from './utilities/store';
 import App from './App';
 import { getBaseName } from '@redhat-cloud-services/frontend-components-utilities/helpers';
+import { IntlProvider } from 'react-intl';
+import messages from './locales/data.json';
```
aneelac22, line 8: should be referencing the locales outside src to avoid extra step of updating another data.json

src/locales/data.json
Hyperkid123: I think it is needed before we get the memsource to create it for us. I think in the first step we will only get one file per language

src/AppEntry.js
Hyperkid123: We should not be referencing resources outside of the `src` folder. Ideally, we would only have one `locales` folder and it would be in the `src` directory.

src/locales/data.json
aneelac22: In that case shouldn't this be auto-generated too? And I think, if not this only works if english is the only language

src/AppEntry.js
aneelac22: I guess the location should be in src but yeah shouldn't there be only one data.json which should be referenced everywhere and it should be the one generated from createDataJson which has the language objects

src/locales/data.json
aneelac22: This already seems to be out of sync with the auto-generated data.json there are 266 strings there and 259 here

src/AppEntry.js
aneelac22: There is also IntlProvider and IntlHelper from frontend-components/translations which does the processing of messages[language] with cases where the translation is not available

package.json
```diff
@@ -114,6 +115,10 @@
     "start:beta": "NODE_ENV=development BETA=true webpack serve --config config/dev.webpack.config.js",
     "start:proxy": "PROXY=true npm run start",
     "start:proxy:beta": "BETA=true npm run start:proxy",
+    "translations": "npm-run-all translations:*",
+    "translations:extract": "npx formatjs extract ./src/*.js --out-file ./locales/translation-template.json",
+    "translations:compile": "npx formatjs compile ./locales/translation-template.json --out-file ./src/locales/translations.json",
```
Hyperkid123, line 120: A note for a future, we will have to change this script to work for multiple files

Hyperkid123 (approved):

PR #1201, opened 2022-08-04
Buttons rendering fixed in review step

victoria-dos:
Buttons render correctly when one creates a new role in the same wizard for the second time.

https://issues.redhat.com/browse/RHCLOUD-20364


fhlavac (approved):
Works great! Thanks @victoria-dos :tada:

PR #1217, opened 2022-08-25
Fix default group ID in requests

fhlavac:
Fixes https://issues.redhat.com/browse/RHCLOUD-20961

- fixed passing a real default group uuid in the API requests to avoid errors
- also added missing translations
- removed unwanted usage of the `useIntl` hook


arivepr (approved):
This looks great! 

I'll obv let @aneelac22 have the final gogo before 🚀

aneelac22 (approved):
LGTM

PR #1256, opened 2022-10-05
Refactor users-list, simplify code

victoria-dos:
Now communicating with redux store only via hooks
Fixed bug "status filters not applied when reloading the page"

### DIFF
--- a/src/helpers/user/user-helper.js
+++ b/src/helpers/user/user-helper.js
@@ -6,11 +6,17 @@ const principalApi = getPrincipalApi();
 const principalStatusApiMap = {
   Active: 'enabled',
   Inactive: 'disabled',
+  All: 'all',
 };
 export async function fetchUsers({ limit, offset = 0, orderBy, filters = {}, inModal, matchCriteria = 'partial' }) {
   const { username, email, status = [] } = filters;
   const sortOrder = orderBy === '-username' ? 'desc' : 'asc';
-  const mappedStatus = status.length === 2 ? 'all' : principalStatusApiMap[status[0]] || 'all';
+  const mappedStatus =
+    typeof status === 'string'
+      ? principalStatusApiMap[status]
+      : status.length === 2
+      ? principalStatusApiMap.All
+      : principalStatusApiMap[status[0]] || principalStatusApiMap.All;
   const response = await principalApi.listPrincipals(limit, offset, matchCriteria, username, sortOrder, email, mappedStatus);
   const isPaginationValid = isOffsetValid(offset, response?.meta?.count);
   offset = isPaginationValid ? offset : getLastPageOffset(response.meta.count, limit);
--- a/src/smart-components/group/add-group/users-list.js
+++ b/src/smart-components/group/add-group/users-list.js
@@ -1,16 +1,14 @@
-import React, { useEffect, Fragment, useState, useContext, useRef } from 'react';
-import { connect, useSelector } from 'react-redux';
+import React, { useEffect, Fragment, useState, useContext, useRef, useCallback } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
 import PropTypes from 'prop-types';
 import { Link, useHistory } from 'react-router-dom';
 import { mappedProps } from '../../../helpers/shared/helpers';
-import { TableToolbarViewOld } from '../../../presentational-components/shared/table-toolbar-view-old';
+import { TableToolbarView } from '../../../presentational-components/shared/table-toolbar-view';
 import { fetchUsers, updateUsersFilters } from '../../../redux/actions/user-actions';
-import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/';
 import { Label } from '@patternfly/react-core';
 import { sortable, nowrap } from '@patternfly/react-table';
 import UsersRow from '../../../presentational-components/shared/UsersRow';
 import {
-  defaultCompactSettings,
   defaultSettings,
   defaultAdminSettings,
   syncDefaultPaginationWithUrl,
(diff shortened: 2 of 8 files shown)


src/smart-components/group/add-group/users-list.js
```diff
@@ -1,82 +1,96 @@
...
     redirected: !inModal && users.pagination.redirected,
   }));
-  const intl = useIntl();
-  const history = useHistory();
+  const selector = ({ userReducer: { users, isUserDataLoading } }) => ({
+    users: users.data && users.data.map((data) => ({ ...data, uuid: data.username })),
+    pagination: users.meta,
+    isLoading: isUserDataLoading,
+  });
+  const { users, pagination, isLoading } = useSelector(selector);
```
karelhala, line 76: I'd rather split this into multiple selectors, that way when for instance just `isUserDataLoading` is changed we will react to it.
```suggestion
  const user = useSelector(({ userReducer: { user } }) => users.data?.map?.((data) => ({ ...data, uuid: data.username }));
  const pagination = useSelector(({ userReducer: { user: { meta } } }) => meta);
  const isLoading = useSelector(({ userReducer: { isUserDataLoading } }) => isUserDataLoading);
```

src/smart-components/group/add-group/users-list.js
```diff
@@ -1,82 +1,96 @@
...
+    ({
+      userReducer: {
+        users: { filters },
+      },
+    }) => (history.location.search.length > 0 || Object.keys(filters).length > 0 ? filters : { status: ['Active'] })
+  );
+
+  const fetchData = (apiProps) => {
+    return dispatch(fetchUsers(apiProps));
+  };
```
karelhala, line 88: Since this won't be changed during re-renders I would change this to a `useCallback`
```suggestion
  const fetchData = useCallback((apiProps) => {
    return dispatch(fetchUsers(apiProps));
  }, [dispatch]);
```

src/smart-components/group/add-group/users-list.js
```diff
@@ -1,82 +1,96 @@
...
+  const fetchUsersFilters = (filters) => {
+    return dispatch(updateUsersFilters(filters));
+  };
```
karelhala, line 91: Same as above, this won't change during re-renders so we don't have to re-calculate it
```suggestion
  const fetchUsersFilters = useCallback((filters) => {
    return dispatch(updateUsersFilters(filters));
  }, [dispatch]);
```

karelhala (commented):
Overall looking good! Just a few cosmetic changes and we are golden.

src/helpers/user/user-helper.js
```diff
@@ -6,11 +6,13 @@ const principalApi = getPrincipalApi();
...
   Active: 'enabled',
   Inactive: 'disabled',
+  All: 'all',
 };
 export async function fetchUsers({ limit, offset = 0, orderBy, filters = {}, inModal, matchCriteria = 'partial' }) {
   const { username, email, status = [] } = filters;
   const sortOrder = orderBy === '-username' ? 'desc' : 'asc';
-  const mappedStatus = status.length === 2 ? 'all' : principalStatusApiMap[status[0]] || 'all';
+  const mappedStatus =
+    typeof status === 'string' ? principalStatusApiMap[status] : status.length === 2 ? 'all' : principalStatusApiMap[status[0]] || 'all';
```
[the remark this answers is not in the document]
victoria-dos: I think both approaches are complex
but
In the first one, we see at first glance that type of _status_ could be a string or list
well, at least for me, the first approach is better for understanding

src/smart-components/group/add-group/users-list.js
```diff
@@ -22,66 +21,112 @@ import { CheckIcon, CloseIcon } from '@patternfly/react-icons';
...
-                      <CheckIcon key="yes-icon" className="pf-u-mr-sm" />
-                      <span key="yes">{intl.formatMessage(messages.yes)}</span>
-                    </span>
-                  </Fragment>
-                ) : (
-                  <Fragment>
-                    <CloseIcon key="no-icon" className="pf-u-mr-sm" />
-                    <span key="no">{intl.formatMessage(messages.no)}</span>
-                  </Fragment>
+// const createRows =
```
fhlavac, line 24: No need for this comment

src/smart-components/group/add-group/users-list.js
```diff
@@ -22,66 +21,112 @@ import { CheckIcon, CloseIcon } from '@patternfly/react-icons';
...
+            selected: Boolean(checkedRows && checkedRows.find((row) => row.uuid === username)),
+          },
+        ],
+        []
+      )
+    : [];
+};
 
-const UsersList = ({ users, fetchUsers, updateUsersFilters, isLoading, pagination, selectedUsers, setSelectedUsers, userLinks, inModal, props }) => {
+// const UsersList = ({ users, fetchUsers, updateUsersFilters, isLoading, pagination, selectedUsers, setSelectedUsers, userLinks, inModal, props }) => {
```
fhlavac, line 92: extra commented code

src/smart-components/group/add-group/users-list.js
```diff
@@ -136,23 +174,32 @@ const UsersList = ({ users, fetchUsers, updateUsersFilters, isLoading, paginatio
...
+        setSortByState({ index, direction });
+        fetchData({ ...pagination, filters, inModal, orderBy });
+      }}
       data={users}
       ouiaId="users-table"
       fetchData={(config) => {
         const status = Object.prototype.hasOwnProperty.call(config, 'status') ? config.status : filters.status;
         const { username, email, count, limit, offset, orderBy } = config;
-        fetchUsers({ ...mappedProps({ count, limit, offset, orderBy, filters: { username, email, status } }), inModal }).then(() => {
+        // fetchUsers({ ...mappedProps({ count, limit, offset, orderBy, filters: { username, email, status } }), inModal }).then(() => {
```
fhlavac, line 198: ^^

src/helpers/user/user-helper.js
fhlavac, line 15: nitpick - how about using `principalStatusApiMap.All` rather than hardcoded `all`
it's probably safer

fhlavac:
Tested locally and everything seems to work fine, also thanks for fixing the status filter bug.
When the comments are removed, we can merge this :tada:

Hyperkid123 (approved):

PR #1271, opened 2022-10-18
Send ref to the roles table names text filter.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-21429
requires: https://github.com/RedHatInsights/frontend-components/pull/1634

### Preview

focus


src/presentational-components/shared/toolbar.js
```diff
@@ -154,7 +160,12 @@ export const filterConfigBuilder = (
                       name: value,
                       orderBy: sortBy,
                     })
-                  );
+                  ).then((data) => {
+                    if (textFilterRef) {
```
karelhala, line 164: Do we need to have the `if` in here as well? Conditional chaining should take care of it.

Hyperkid123: (I want to kill conditional chaining :smiling_imp: )

karelhala (approved):

PR #1282, opened 2022-10-25
added displayName check

JuliePrazakova:
https://issues.redhat.com/browse/RHCLOUD-21426


config/dev.webpack.config.js
```diff
@@ -13,7 +13,7 @@ const webpackProxy = {
   deployment: process.env.BETA ? 'beta/apps' : 'apps',
   useProxy: true,
   proxyVerbose: true,
-  env: 'stage-beta',
+  env: 'prod-beta',
```
fhlavac, line 16: we probably don't need to commit that

src/smart-components/role/add-role-new/validators.js
```diff
@@ -19,7 +19,13 @@ export const asyncValidator = async (roleName) => {
     return undefined;
   });
 
-  if (response?.data?.length > 0) {
+  //it has to be here twice because api is using and instead of or
```
fhlavac, line 22:
```suggestion
  // it has to be here twice because api is using and instead of or
```

src/test/smart-components/addRoleWizard/validators.test.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  test('should throw nameAlreadyTaken message', async () => {
+    fetchRoles.mockImplementationOnce(() => Promise.resolve({ data: [1] }));
+    fetchRoles.mockImplementationOnce(() => Promise.resolve({ data: [] }));
+
+    expect.assertions(1);
+    await expect(asyncValidator('foo')).rejects.toEqual('Name has already been taken.');
+  });
+  test('should throw error and return undefined', async () => {
```
fhlavac, line 41:
```suggestion

  test('should throw error and return undefined', async () => {
```

fhlavac:
Just two unused imports and some nitpicking. Otherwise looks good!

src/smart-components/role/add-role-new/validators.js
```diff
@@ -19,7 +19,13 @@ export const asyncValidator = async (roleName) => {
     return undefined;
   });
 
-  if (response?.data?.length > 0) {
+  //it has to be here twice because API is using AND instead of OR
```
fhlavac, line 22:
```suggestion
  // it has to be here twice because API is using AND instead of OR
```

fhlavac (approved):

PR #1351, opened 2023-01-20
Make Roles table expandable

fhlavac:
[RHCLOUD-21025](https://issues.redhat.com/browse/RHCLOUD-21025)

- made Roles table expandable
- enhanced tests

chrome-capture-2023-0-20 (2)


src/smart-components/role/roles.scss
```diff
@@ -1,12 +1,15 @@
...
+    --pf-c-button--hover--after--BorderWidth: 0;
+  }
 
-   .pf-c-table tbody tr.rbac-c-role-default .pf-c-table__check label input:hover {
+  .pf-c-table tbody tr.rbac-c-role-default .pf-c-table__check label input:hover {
     cursor: not-allowed;
-   }
+  }
+  .pf-c-table__compound-expansion-toggle {
+    outline: none;
```
Hyperkid123, line 13: We should not be disabling the outline. It has its purpose although it might look bad in some cases.

Hyperkid123 (approved):

PR #1355, opened 2023-01-30
Fix add-permissions wizard

fhlavac:
Fixes [RHCLOUD-23991](https://issues.redhat.com/browse/RHCLOUD-23991)

cc @jjaquish

### DIFF
--- a/src/smart-components/role/add-role-permissions/add-role-permission-wizard.js
+++ b/src/smart-components/role/add-role-permissions/add-role-permission-wizard.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef, createContext } from 'react';
+import React, { useState, useEffect, useRef, createContext, useMemo } from 'react';
 import { useDispatch } from 'react-redux';
 import AddRolePermissionSummaryContent from './add-role-permissions-summary-content';
 import AddRolePermissionSuccess from './add-role-permission-success';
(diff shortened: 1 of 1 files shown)


src/smart-components/role/add-role-permissions/add-role-permission-wizard.js
```diff
@@ -47,10 +46,7 @@ const AddRolePermissionWizard = ({ role }) => {
   const setWizardError = (error) => setWizardContextValue((prev) => ({ ...prev, error }));
   const setWizardSuccess = (success) => setWizardContextValue((prev) => ({ ...prev, success }));
   const setHideForm = (hideForm) => setWizardContextValue((prev) => ({ ...prev, hideForm }));
-
-  useEffect(() => {
-    setSchema(schemaBuilder(container.current));
-  }, []);
+  const schema = useMemo(() => schemaBuilder(container.current));
```
Hyperkid123, line 49: You want to add a trigger to the useMemo so it does not computer on each render.

```js
const schema = useMemo(() => schemaBuilder(container.current), []);
```

Hyperkid123 (changes requested):

src/smart-components/role/add-role-permissions/add-role-permission-wizard.js
fhlavac: ah, forgot to add that, thank you. Fixed

Hyperkid123 (approved):

PR #1373, opened 2023-03-08
Remove mapStateToProps and mapDispatchToProps

apinkert:
For [RHCLOUD-24543](https://issues.redhat.com/browse/RHCLOUD-24543)

### DIFF
--- a/src/smart-components/group/edit-group-modal.js
+++ b/src/smart-components/group/edit-group-modal.js
@@ -13,7 +13,7 @@ import { debouncedAsyncValidator } from './validators';
 import { useIntl } from 'react-intl';
 import messages from '../../Messages';
 import pathnames from '../../utilities/pathnames';
-import { useDispatch, useSelector } from 'react-redux';
+import { useDispatch } from 'react-redux';
 
 const EditGroupModal = ({ postMethod, pagination, filters, cancelRoute, submitRoute = cancelRoute, group, onClose }) => {
   const intl = useIntl();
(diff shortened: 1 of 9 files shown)


src/smart-components/group/edit-group-modal.js
```diff
@@ -44,6 +33,12 @@ const EditGroupModal = ({
         .catch(() => setGroupData(undefined));
   };
 
+  useSelector(({ groupReducer: { isLoading } }) => ({
```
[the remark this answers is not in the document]
apinkert: Removed these two selectors as `isLoading` was never being used in this component

src/smart-components/group/principal/add-group-members.js
```diff
@@ -1,57 +1,55 @@
...
-  closeUrl,
-  addMembersToGroup,
-  fetchMembersForGroup,
-  fetchGroups,
-}) => {
+const AddGroupMembers = ({ closeUrl }) => {
   const [selectedUsers, setSelectedUsers] = useState([]);
   const intl = useIntl();
+  const { push } = useHistory();
+  const uuid = useRouteMatch('/groups/detail/:uuid') ? useRouteMatch('/groups/detail/:uuid').params.uuid : null;
```
apinkert, line 17: @Hyperkid123 Not 100% sure on this one, but I was having reference issues when I left `match` above. This fixed those reference issues as well as the tests that were failing. I reference how `useRouteMatch()` was being used in `edit-group-modal.js`

Hyperkid123: You can get the `match` from `useRouteMatch`. But you can't have hooks in conditions. You want to get the match object and then attempt to find the `uuid`.
```diff
- const uuid = useRouteMatch('/groups/detail/:uuid') ? useRouteMatch('/groups/detail/:uuid').params.uuid : null;
+ const match = useRouteMatch('/groups/detail/:uuid')
+ const uuid = match?.params.uuid
```
But I would consider using `useParams` hook and look foe the `uuid` param there. That hook should give you all params from URl. But if the route mask is not correct it might not work.

Hyperkid123 (approved):

PR #1376, opened 2023-03-16
Upgrade react-router to v6

fhlavac:
[RHCLOUD-24537](https://issues.redhat.com/browse/RHCLOUD-24537)

- migrated RBAC to use React Router v6
- did some code cleaning
- renamed `inModal` param used for persisting filters and search in URL query to `usesMetaInURL`
- simplified a bit process of persisting query params in the URL


src/entries/MyUserAccess.js
```diff
@@ -18,31 +19,29 @@ import '../App.scss';
...
+const MuaApp = () => {
+  const chrome = useChrome();
+  const navigate = useAppNavigate('/iam');
+  const userData = useUserData();
+
   useEffect(() => {
-    // redirect to MUA if url is "/settings"
-    if (window?.location?.pathname?.match(/\/(iam|settings)$/)) {
-      history.push('/my-user-access');
+    if (window?.location?.pathname?.match(/\/(iam)$/)) {
```
Hyperkid123, line 28: We no longer need the window pathname reference we can use the location pathname as the history is shared with chrome.

src/entries/MyUserAccess.js
```diff
@@ -18,31 +19,29 @@ import '../App.scss';
...
-const MuaApp = () => {
-  const userData = useUserData();
   return (
     <PermissionsContext.Provider value={{ ...userData }}>
       <ErroReducerCatcher>
         <section style={{ marginLeft: 0, padding: 0 }}>
-          <Routes />
+          <Suspense fallback={<AppPlaceholder />}>
+            <Routes>
+              <Route path={'/'} element={<MyUserAccess />} />
```
Hyperkid123, line 42: the path can be just `path="/"`

PR #1400, opened 2023-03-29
Creating AddUserToGroup test suite

arivepr:
This PR addresses [this JIRA card](https://issues.redhat.com/browse/RHCLOUD-23769)
Changes include:

- Creating a testing suite for the AddUser modal for:
    - Rendering the modal
    - Correct actions being called at the right time
    - Basic interaction of saving the user
    - Blocking the user from adding to group if not an admin
- Minimal changes for test facilitation
- Minimal changes to outside failing tests


fhlavac:
@arivepr it would be also nice to add a test suite for the Users page checking if the "Add user to group" button is visible only for orgAdmin or userAccessAdministrator.

src/test/smart-components/user/addUserToGroup.test.js
```diff
@@ -0,0 +1,115 @@
...
+      }),
+    }));
+   
+  });
+
+  test('Add to group wizard renders correctly', () => {
+    const store = mockStore(initialState);
+
+    renderComponent('testuser', store);
+    expect(screen.getByText(/Add to group/i)).toBeInTheDocument();
```
fhlavac, line 85: Can we also check the warning "This group list has been filtered to only show groups that the user is not currently a member of." is visible? We can directly reference the intl message

fhlavac: I think we can test both :slightly_smiling_face:

arivepr:
@fhlavac So looking everything over, I agree on adding a test case to see if the user is able to add themselves to a group. However given that we're switching over RBAC testing to RTL and it's a different component, I think that would like outside the scope of this particular ticket.

Would you agree to me opening up a separate ticket for that, in order to not bloat this one with adding the test case/migrating to RTL in User, and then pull that once I free up cycles this sprint, if not next?

For this one I am instead adding a test to make sure user can't select or add groups if they aren't an admin 😄

fhlavac:
@arivepr good point, in this case, I think we can update [RHCLOUD-23749](https://issues.redhat.com/browse/RHCLOUD-23749) with this requirement. I'll do that.

src/test/smart-components/user/addUserToGroup.test.js
```diff
@@ -0,0 +1,136 @@
...
+    expect(fetchGroups).toHaveBeenCalled();
+  });
+
+  test('User groups table is rendered, with fetch group action dispatched', async () => {
+    renderComponent('testUser', store);
+
+    expect(fetchGroups).toHaveBeenCalled();
+    await waitFor(() => expect(screen.getByLabelText('groups table')));
+    await waitFor(() => expect(screen.getByText(`${messages.onlyNonUserGroupsVisible.defaultMessage}`)));
+    screen.debug();
```
fhlavac, line 107: please remove the `screen.debug()`

src/test/smart-components/user/addUserToGroup.test.js
```diff
@@ -0,0 +1,136 @@
...
+      }),
+    }));
+    renderComponent('testUser', store);
+
+    await waitFor(() => expect(screen.getByLabelText(`group-name-${testGroups[0].uuid}`)));
+    userEvent.click(screen.getByLabelText(`Select row 0`));
+    userEvent.click(screen.getByLabelText(`Save`));
+
+    await waitFor(() => expect(addMembersToGroup).toHaveBeenCalled());
+    screen.debug();
```
fhlavac, line 134: please remove the `screen.debug()`

fhlavac:
@arivepr we should have the test file name the same as its tested file. I know the component files should be named in CamelCase but we can probably change it for all files and tests in one PR to make that consistent

src/test/smart-components/user/addUserToGroup.test.js
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { render, screen, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import AddUserToGroup from '../../../smart-components/user/add-user-to-group/add-user-to-group';
```
fhlavac, line 5: Just nitpicking, it's more clear to have "local imports" below the ones from external libraries

fhlavac:
image

we are not testing displaying a cancel warning

...and also making sure that filtering works
image

Hyperkid123 (approved):

PR #1401, opened 2023-03-30
Refactor groups, remove TableToolbarViewOld

victoria-dos:


src/smart-components/myUserAccess/MUAAccessTable.js
```diff
@@ -42,18 +41,41 @@ const MUAAccessTable = ({ filters, setFilters, apps, hasActiveFilters, showResou
     });
 
   useEffect(() => {
-    fetchData(defaultSettings);
+    fetchData({ limit: 20, offset: 0, itemCount: 0, orderBy });
```
fhlavac, line 44: can we preserve using `defaultSettings` object as a base?

fhlavac:
@victoria-dos just one nitpick. Otherwise looks good to me :slightly_smiling_face:

PR #1409, opened 2023-04-11
Changes to RBAC UI for OCM Roles

coderbydesign:
**Jira:** https://issues.redhat.com/browse/RHCLOUD-25197

There are a handful of changes required to the RBAC UI for "My User Access" now that OCM roles can be assigned to groups in order for users to view their OCM roles/permissions in MUA:

- [x] The alert showing "OpenShift Clusters permissions are not managed with User Access" should now be removed from the OpenShift bundle when displaying OpenShift roles/permissions in MUA. Similar to what's used in the Ansible bundle, defaulting to the RHEL implementation of `CommonBundleView`.
- [x] "ocm" should be added to the dropdown for the role application filter in MUA for the OpenShift bundle.

**TODO:** Since `ocm` roles do not contain permissions, the queries to `fetchRoles` on page load, and when the filter is updated, will still not return OCM roles, because `?application=<application>` is based on the application name in the permissions in the RBAC API. 

~There are two potential solutions for this:
Update the UI code to query `?external_tenant=ocm` in the case where OCM is being filtered, which _would_ return the OCM role(s). One problem there is that the `?application=` filter would not work when combined with `?external_tenant=` because it would look for _both_ of those filters, potentially requiring multiple API calls. The config for `appsIds` in `bundles.js` would also need to be updated to support this.~
The solution will be to push this logic into the RBAC API so that `?application=ocm` would still work for external tenant roles, falling back to check `external_tenant` when there are no permissions matching the `application` param value. There may be performance and other implications here that would need to be thought through more. Supporting PR: https://github.com/RedHatInsights/insights-rbac/pull/853

- [x] Selecting "ocm" from the application filter dropdown should return OCM roles, alone or in combination with other application filters.
- [x] Roles containing "OCM" in the role name should be returned when filtering roles by name in MUA for the OpenShift bundle.

---- 
Screen Shot 2023-04-10 at 4 35 39 PM

Screen Shot 2023-04-11 at 7 50 49 AM


src/smart-components/myUserAccess/bundles/openshift.js
```diff
@@ -1,23 +1 @@
...
-
-import './MUABundles.scss';
-
-const OpenshiftBundle = (props) => {
-  const intl = useIntl();
-  return (
-    <Fragment>
-      <div className="pf-u-mb-xl">
-        <Alert isInline variant="info" title="OpenShift Clusters permissions are not managed with User Access">
-          <p>{intl.formatMessage(messages.openshiftPermissions)}</p>
```
Hyperkid123, line 15: Can you remove this message from `src.Messages.js`? Seems like it's not used anywhere else.

Hyperkid123 (commented):
Works in UI, just a small nitpick.

src/smart-components/myUserAccess/bundles/openshift.js
coderbydesign: @Hyperkid123 done!

Hyperkid123 (approved):

PR #1426, opened 2023-05-02
Migrate to TS, use TableComposable for Users page (not-selectable users-list)

victoria-dos:
[RHCLOUD-25631](https://issues.redhat.com/browse/RHCLOUD-25631)

- Create the replacement for the current TableToolbarView (migrate to TS)
- Inside this new component create a new Table component using PF TableComposable (instead of legacy Table)
- Apply changes to the Users page (only for the non-selectable Users list)


src/presentational-components/shared/table-composable-toolbar-view.tsx
```diff
@@ -0,0 +1,272 @@
...
+    { title: string | React.RefAttributes<HTMLAnchorElement> }, //link to user or just username
+    string, //email
+    string, //firstName
+    string, //lastName
+    { title: React.FC<LabelProps>; props: { 'data-is-active': boolean } }
+  ];
+  selected: boolean;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Hyperkid123, line 28: Is this legacy or future type? If the type is not used, we don't want it to be defined.

src/presentational-components/shared/table-composable-toolbar-view.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+  const renderEmpty = () => (
+    <EmptyWithAction
+      title={intl.formatMessage(messages.noMatchingItemsFound, { items: title.plural })}
+      description={
+        noData && noDataDescription
+          ? noDataDescription
+          : [intl.formatMessage(messages.filterMatchesNoItems, { items: title.plural }), intl.formatMessage(messages.tryChangingFilters)]
+      }
+      icon={false} //for types
```
Hyperkid123, line 153: We should make the prop optional

src/smart-components/user/user-table-helpers.js
Hyperkid123, line 1: Any reason why this new file is not `.tsx`?

src/smart-components/user/users-list-not-selectable.js
Hyperkid123, line 1: Same here, new files should be written in TS

src/presentational-components/shared/toolbar.js
```diff
@@ -7,9 +7,9 @@ import { pickBy } from 'lodash';
 import { selectedRows, calculateChecked, debouncedFetch, firstUpperCase } from '../../helpers/shared/helpers';
 import { calculateOffset, calculatePage, defaultSettings } from '../../helpers/shared/pagination';
 
-export const paginationBuilder = (pagination = {}, fetchData = () => undefined, filterValue = '', sortBy = '') => ({
+export const paginationBuilder = (pagination = {}, fetchData, filterValue = '', sortBy = '') => ({
   ...pagination,
-  itemCount: pagination.count,
+  count: pagination.count,
```
fhlavac, line 12: this needs to be reverted. Pagination PF component needs the property to be named like this

victoria-dos:
@Hyperkid123 @fhlavac

src/helpers/shared/pagination.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (!defaultPagination) {
+    defaultPagination = defaultSettings;
+  }
+  const searchParams = new URLSearchParams(location.search);
+
+  let limit = parseInt(searchParams.get('per_page') as string);
+  let page = parseInt(searchParams.get('page') as string);
+
+  if (isNaN(limit) || limit <= 0) {
+    limit = defaultPagination.limit as number;
```
Hyperkid123, line 41: I don't think we need the `as number` now that the default object has all attributes required.

src/presentational-components/shared/empty-state.tsx
```diff
@@ -1,16 +1,23 @@
...
+
+const EmptyWithAction = ({ title, icon, description, actions, ...props }: EmptyWithActionProps) => (
   <EmptyState variant={EmptyStateVariant.small} {...props}>
     <EmptyStateIcon icon={icon || SearchIcon} />
     <Title headingLevel="h4" size="lg">
       {title}
     </Title>
     <EmptyStateBody className="pf-u-mb-md">
-      {description.map((text, key) => (
+      {description.map((text: any, key: number) => (
```
Hyperkid123, line 20: I think the `text` should be a `React.ReactNode` type.

src/presentational-components/shared/table-composable-toolbar-view.tsx
```diff
@@ -0,0 +1,260 @@
...
+          icon={PlusCircleIcon}
+          description={[
+            intl.formatMessage(messages.toConfigureUserAccess),
+            intl.formatMessage(messages.createAtLeastOneItem, { item: title.singular }),
+          ]}
+          actions={toolbarButtons ? toolbarButtons()[0] : false}
+          {...(typeof emptyProps === 'object' ? emptyProps : {})}
+        />
+      ) : (
+        renderTable()
```
Hyperkid123, line 256: In general we should move away from these render functions. It can cause trouble in React 18. We have two options:

1. Move the entire JSX and remove the wrapping render functions.
2. Create new component and pass all relevant props to it.

src/presentational-components/shared/toolbar.js
```diff
@@ -7,7 +7,7 @@ import { pickBy } from 'lodash';
 import { selectedRows, calculateChecked, debouncedFetch, firstUpperCase } from '../../helpers/shared/helpers';
 import { calculateOffset, calculatePage, defaultSettings } from '../../helpers/shared/pagination';
 
-export const paginationBuilder = (pagination = {}, fetchData = () => undefined, filterValue = '', sortBy = '') => ({
+export const paginationBuilder = (pagination = {}, fetchData, filterValue = '', sortBy = '') => ({
```
Hyperkid123, line 10: The `fetchData` prop should now be required if the default was removed.

src/presentational-components/shared/table-composable-toolbar-view.tsx
victoria-dos: That is a good point! thanks

fhlavac:
:eyes:

src/presentational-components/shared/table-composable-toolbar-view.tsx
```diff
@@ -0,0 +1,367 @@
...
+    type?: any;
+    items?: any;
+  }>;
+  isFilterable?: boolean;
+  onShowMore?: () => void;
+  showMoreTitle?: string;
+  onFilter?: () => void;
+  onChange?: () => void;
+  value?: unknown;
+  sortBy: ISortBy; // { index: number; direction: 'asc' | 'desc'; defaultDirection?: 'asc' | 'desc'; }
```
fhlavac, line 87: this comment is probably not needed

src/smart-components/common/page-action-route.js
```diff
@@ -4,7 +4,9 @@ import PropTypes from 'prop-types';
 const PageActionRoute = ({ pageAction, children }) => {
   useEffect(() => {
     insights.chrome.appAction(pageAction);
-    return () => insights.chrome.appAction(undefined);
+    return () => {
```
fhlavac, line 7: nitpick - not needed change

src/smart-components/user/user-table-helpers.tsx
```diff
@@ -0,0 +1,68 @@
...
+  is_active: boolean;
+  is_org_admin: boolean;
+  last_name: string;
+  username: string;
+  uuid: string;
+}
+
+export type CellObject = { title: string | React.RefAttributes<HTMLAnchorElement>; props?: { 'data-is-active': boolean } };
+
+export interface RowProps {
```
fhlavac, line 20: I see you use this interface to define `TableProps`. We should come up with a more generic solution that would match different data passed to the table.

src/smart-components/user/user-table-helpers.tsx
```diff
@@ -0,0 +1,68 @@
...
+  uuid: string; //username
+  cells: [
+    React.ReactNode, //yes or no for isOrgAdmin
+    CellObject, //link to user or just username
+    string, //email
+    string, //firstName
+    string, //lastName
+    CellObject //status
+  ];
```
fhlavac, line 29:
```suggestion
  uuid: string; // username
  cells: [
    React.ReactNode, // yes or no for isOrgAdmin
    CellObject, // link to user or just username
    string, // email
    string, // firstName
    string, // lastName
    CellObject // status
  ];
```

src/smart-components/user/users.tsx
```diff
@@ -33,13 +33,13 @@ const Users = () => {
...
         <Section type="content" id="users">
-          <UsersList
+          <UsersListNotSelectable
             userLinks={activeUserPermissions.userAccessAdministrator || activeUserPermissions.orgAdmin}
             props={{
               isSelectable: false,
               isCompact: false,
             }}
-            usesMetaInURL
+            usesMetaInURL={true}
```
fhlavac, line 42: not needed change

src/presentational-components/shared/table-composable-toolbar-view.tsx
```diff
@@ -0,0 +1,367 @@
...
+
+// TODO: make it clear later and remove FetchDataProps from below
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface FetchDataPropsNew {
+  pagination?: PaginationProps;
+  filters?: FilterProps;
+  orderBy?: string;
+}
+
+interface FetchDataProps {
```
fhlavac, line 36: ALso here we could extend PaginationProps

src/presentational-components/shared/table-composable-toolbar-view.tsx
```diff
@@ -0,0 +1,367 @@
...
+interface PaginationProps {
+  limit?: number;
+  offset?: number;
+  count?: number;
+  noBottom?: boolean;
+}
+
+// TODO: make it clear later and remove FetchDataProps from below
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface FetchDataPropsNew {
```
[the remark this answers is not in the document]
victoria-dos: I think I will remove it, at first I wanted it for some clarity and order, but it is not necessary any more since FetchDataProps is OK

src/smart-components/user/user-table-helpers.tsx
victoria-dos: I understand, changing Table prop into Array<unknown> for now

src/presentational-components/shared/table-composable-toolbar-view.tsx
victoria-dos: I removed PaginationProps for now, it was not used

src/helpers/shared/pagination.js
```diff
@@ -53,8 +56,10 @@ export const isPaginationPresentInUrl = (location) => {
 export const isOffsetValid = (offset = 0, count = 0) => offset === 0 || count > offset;
 
 export const getLastPageOffset = (count, limit) => Math.floor((count % limit === 0 ? count - 1 : count) / limit) * limit;
-
-export const applyPaginationToUrl = (location, navigate, limit, offset = 0) => {
+export const applyPaginationToUrl = (location, navigate, limit, offset) => {
```
fhlavac: this change is not needed

fhlavac (approved):

PR #1490, opened 2023-07-20
RHCLOUD-25457 RBAC | Group Explorer View

aferd:
:warning: WIP


src/smart-components/group/group-table-helpers.js
```diff
@@ -40,48 +42,144 @@ DefaultPlatformPopover.propTypes = {
...
+          },
         ],
-        selected: Boolean(selectedRows && selectedRows.find((row) => row.uuid === uuid)),
       },
-    ],
+    ]),
     []
   );
 };
+// cells: [
```
fhlavac, line 156: it looks like the empty initial array for reduce is in the wrong position

```suggestion
    ], [])
  );
};
```

aferd: ah, thank you!

PR #1503, opened 2023-08-02
Update master-stable with latest RBAC UI

florkbr:
Suspect we may have fixed some bugs in master that we're seeing on stage-stable (does not occur on stage-preview). See thread: https://redhat-internal.slack.com/archives/C0235KZ3Q0J/p1690991086476359


florkbr:
@fhlavac is there anything in master that's not ready for release?

fhlavac (approved):

fhlavac:
Nope, we can release everything 🙂

PR #1505, opened 2023-08-07
Unit tests for inv group roles - DRAFT 

arivepr:
Dropping this here for visibility. Hoping to get some eyes on whatever it is I might be doing wrong. 

Components seem to not be rendering when trying to test for DDF related connection/things, might be where I'm messing up.


src/redux/reducers/inventory-reducer.js
```diff
@@ -12,7 +12,7 @@ const setResourceTypes = (state, { payload, meta }) => {
   const data = payload.data.reduce((acc, curr) => ({ ...acc, [curr.name]: { ...curr } }), {});
   return {
     ...state,
-    resourceTypes: meta.permissions.reduce(
+    resourceTypes: meta.permissions?.reduce(
```
fhlavac, line 15: this does not seem to be needed

src/test/smart-components/addRoleWizard/inventory-groups-role.test.js
```diff
@@ -0,0 +1,157 @@
...
+      type: FETCH_INVENTORY_GROUP,
+      payload: Promise.resolve({}),
+    }));
+
+    const store = mockStore(initialStateWithPermissions);
+    renderComponent(store);
+
+    fireEvent.click(screen.getAllByLabelText('Options menu')[0]);
+
+    expect(screen.getByText('fooBar')).toBeInTheDocument();
```
fhlavac, line 114: here we could also check at least last group visible or number of items (in a follow-up also "See more" visibility could be checked as a different test case)

fhlavac: another thing to check should be the presence and functionality of the "Select all" option

fhlavac: also "Ungrouped systems" option needs to be checked for the host permissions. We'll add it to the follow-up ticket

fhlavac:
@arivepr apart from the comments looks good! It would be nice if you could add the smaller things mentioned and we can merge this PR today. Then we could create a follow-up to add what is missing.

fhlavac (approved):

PR #1515, opened 2023-08-16
RBAC - Group Explorer View

aferd:
RHCLOUD-25457 re-do.

Pagination and sorting will be completed in a follow up ticket [RHCLOUD-29186](https://issues.redhat.com/browse/RHCLOUD-29186)


src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,112 @@ DefaultPlatformPopover.propTypes = {
...
+                  variant={TableVariant.compact}
+                  cells={[
+                    intl.formatMessage(messages.orgAdmin),
+                    intl.formatMessage(messages.username),
+                    intl.formatMessage(messages.email),
+                    intl.formatMessage(messages.firstName),
+                    intl.formatMessage(messages.lastName),
+                    intl.formatMessage(messages.status),
+                  ]}
+                  rows={principalCount.map((member) => ({
```
aferd, line 140: I am still unable to access the right data for both tables.. and on this one specifically it errors after clicking with `Cannot read properties of undefined (reading 'limit')`

src/smart-components/group/groups.js
```diff
@@ -2,6 +2,7 @@ import React, { Suspense, useContext, useEffect, useRef, useState } from 'react'
 import { Outlet, useLocation, useNavigate } from 'react-router-dom';
 import { sortable } from '@patternfly/react-table';
 import { Button, Stack, StackItem } from '@patternfly/react-core';
+import { compoundExpand, nowrap } from '@patternfly/react-table';
```
fhlavac, line 5:
```suggestion
import { Button, Stack, StackItem } from '@patternfly/react-core';
import { compoundExpand, nowrap, sortable } from '@patternfly/react-table';
```

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,112 @@ DefaultPlatformPopover.propTypes = {
...
+        uuid: `${uuid}-roles`,
+        parent: 3 * i,
+        compoundParent: 1,
+        fullWidth: true,
+        noPadding: true,
+        cells: [
+          {
+            props: { colSpan: 7, className: 'pf-m-no-padding' },
+            title:
+              roleCount.length > 0 ? (
```
fhlavac, line 83: here you are calling map on a number, you need to call it on the array of roles

src/smart-components/group/group-table-helpers.js
fhlavac, line 140: here you are calling map on a number, you need to call it on the array of members

src/redux/actions/group-actions.js
```diff
@@ -174,6 +174,7 @@ export const removeMembersFromGroup = (groupId, members) => {
 export const fetchRolesForGroup = (groupId, pagination, options) => ({
```
fhlavac, line 174: There I would either:
- add a param that would switch between fetching roles for a group detail and expandable table in groups
- or have separate action for fetching the data for group detail (stored as selectedGroup in redux) and expanded Groups rows (stored in groups entries)

src/smart-components/group/group-table-helpers.js
aferd: i should be able to pull this information from the data that is being reduced above, correct?

src/redux/actions/group-actions.js
aferd: why do I need to separate the data, even though it is the same data in each group detail and expandable table?

fhlavac: yeah, it's the same data but you probably don't want to store them in `selectedGroup` object in store but rather enhance given group in the `groups.data` array. Otherwise, you would lose previously fetched data on every expand - new group detail data would overwrite the old one.

aferd:
I am having some weird interactions, I am able to click "roles" or "members" on different rows twice and then it errors out with `Cannot read properties of null (reading 'isEnabled')`

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,116 @@ DefaultPlatformPopover.propTypes = {
...
 };
 
-export const createRows = (isAdmin, data, selectedRows = []) => {
+export const createRows = (isAdmin, data, selectedRows, expanded = []) => {
   const intl = useIntl();
   return data.reduce(
-    (acc, { uuid, name, roleCount, principalCount, modified, platform_default: isPlatformDefault, admin_default: isAdminDefault }) => [
+    (
+      acc,
+      { uuid, name, roleCount, principalCount, modified, expandTableDetails, platform_default: isPlatformDefault, admin_default: isAdminDefault },
```
fhlavac, line 50:
```suggestion
      { uuid, name, roleCount, principalCount, modified, roles, members, platform_default: isPlatformDefault, admin_default: isAdminDefault },
```
expandTableDetails does not exist - so far you have the data persisted under `roles` key (members are not added yet)

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,116 @@ DefaultPlatformPopover.propTypes = {
...
+                  ouiaId="roles-in-group-nested-table"
+                  aria-label="Simple Table"
+                  variant={TableVariant.compact}
+                  cells={[
+                    intl.formatMessage(messages.roleName),
+                    intl.formatMessage(messages.description),
+                    intl.formatMessage(messages.permissions),
+                    intl.formatMessage(messages.lastModified),
+                  ]}
+                  rows={expandTableDetails.map((roles) => {
```
fhlavac, line 98:
```suggestion
                  rows={roles.map((role) => {
```

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,116 @@ DefaultPlatformPopover.propTypes = {
...
+                  variant={TableVariant.compact}
+                  cells={[
+                    intl.formatMessage(messages.orgAdmin),
+                    intl.formatMessage(messages.username),
+                    intl.formatMessage(messages.email),
+                    intl.formatMessage(messages.firstName),
+                    intl.formatMessage(messages.lastName),
+                    intl.formatMessage(messages.status),
+                  ]}
+                  rows={expandTableDetails.map((members) => ({
```
fhlavac, line 144:
```suggestion
                  rows={members.map((member) => ({
```
add at least an empty array, without it, there will be still an error

src/smart-components/group/groups.js
```diff
@@ -179,7 +181,19 @@ const Groups = () => {
   const data = groups.map((group) =>
     group.platform_default || group.admin_default ? { ...group, principalCount: `All${group.admin_default ? ' org admins' : ''}` } : group
   );
-  const rows = createRows(isAdmin, data, selectedRows);
+
+  const fetchExpandTableDetailsForGroup = (uuid) => dispatch(fetchRolesForGroup(uuid));
```
fhlavac, line 185: also here it complains about not passed pagination params

src/smart-components/group/groups.js
```diff
@@ -176,10 +179,24 @@ const Groups = () => {
...
       : []),
   ];
+
   const data = groups.map((group) =>
     group.platform_default || group.admin_default ? { ...group, principalCount: `All${group.admin_default ? ' org admins' : ''}` } : group
   );
-  const rows = createRows(isAdmin, data, selectedRows);
+
+  const fetchExpandTableRoleDetailsForGroup = (uuid) => dispatch(fetchRolesForGroup(uuid, defaultCompactSettings));
+  const fetchExpandTableMemberDetailsForGroup = (uuid) => dispatch(fetchMembersForGroup(uuid, defaultCompactSettings));
```
fhlavac, line 188: Can we rename it to something shorter? Like...
```suggestion
  const fetchExpandedRoles = (uuid) => dispatch(fetchRolesForGroup(uuid, defaultCompactSettings));
  const fetchExpandedMembers = (uuid) => dispatch(fetchMembersForGroup(uuid, defaultCompactSettings));
```

src/smart-components/group/groups.js
```diff
@@ -176,10 +179,24 @@ const Groups = () => {
...
+
+  const onExpand = (_event, _rowIndex, colIndex, isOpen, rowData) => {
+    if (!isOpen) {
+      setExpanded({ ...expanded, [rowData.uuid]: colIndex + Number(!isAdmin) });
+      colIndex + Number(!isAdmin) === 2 && fetchExpandTableRoleDetailsForGroup(rowData.uuid) && fetchExpandTableMemberDetailsForGroup(rowData.uuid);
```
fhlavac, line 193: we probably want to fetch roles data for index === 2 and members for index === 3. Now it is fetching both for the roles cell

src/smart-components/group/groups.js
fhlavac, line 188: fetchMembersForGroup should be updated similarly as fetchRolesForGroup to store fetched data in the redux `groups`, not `selectedGroup`

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,134 @@ DefaultPlatformPopover.propTypes = {
...
+                    />
+                  )}
+                </Fragment>
+              </>
+            ) : (
+              name
+            ),
+          },
+          { title: roleCount, props: { isOpen: expanded[uuid] === 2 } },
+          { title: principalCount, props: { isOpen: expanded[uuid] === 3 } },
```
aferd, line 85: here, I am trying to make the members name "read-only" for the default groups - I tried putting a conditional like this ` { title: isPlatformDefault || isAdminDefault ? principalCount : principalCount, props: { isOpen: expanded[uuid] === 3 } }` but it didn't do what I thought it would.

aferd:
Screen Shot 2023-11-01 at 3 37 51 PM
Screen Shot 2023-11-01 at 3 32 50 PM

this is the current state... if you look at the tables, the `members` table is indented more than the `roles `table so I am trying to fix that..

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +44,152 @@ DefaultPlatformPopover.propTypes = {
...
+                        )}
+                        {intl.formatMessage(member?.is_org_admin ? messages.yes : messages.no)}
+                      </TextContent>,
+                      member.first_name,
+                      member.last_name,
+                      member.username,
+                      member.email,
+                      <Label key={member?.is_active} color={member?.is_active && 'green'}>
+                        {intl.formatMessage(member?.is_active ? messages.active : messages.inactive)}
+                      </Label>,
```
fhlavac, line 182: You could use something like `<ListLoader items={3} isCompact />` for the loading state

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +44,152 @@ DefaultPlatformPopover.propTypes = {
...
+                    />
+                  )}
+                </Fragment>
+              </>
+            ) : (
+              name
+            ),
+          },
+          { title: roleCount, props: { isOpen: expanded[uuid] === 2 } },
+          { title: isPlatformDefault || isAdminDefault ? principalCount : principalCount, props: { isOpen: expanded[uuid] === 3 } },
```
fhlavac, line 88: this condition does not seem to make sense

aferd: I thought these were attached to the default groups and therefore this would just show the name without the ability to expand, otherwise make it expandable.

fhlavac: I mean this `isPlatformDefault || isAdminDefault ? principalCount : principalCount` you still return just principalCount

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +44,152 @@ DefaultPlatformPopover.propTypes = {
...
+                <Text className="pf-u-mx-lg pf-u-my-sm">{intl.formatMessage(messages.noGroupRoles)}</Text>
+              ),
+          },
+        ],
+      },
+      {
+        uuid: `${uuid}-members`,
+        parent: 3 * i,
+        compoundParent: 2,
+        fullwidth: true,
```
fhlavac, line 143:
```suggestion
        fullWidth: true,
```

this will fix the width issue

src/smart-components/group/group-table-helpers.js
aferd: will the props not be passed in with the second one? I see that they are not, but I thought they would. Looking at it more.

src/smart-components/group/group-table-helpers.js
```diff
@@ -9,6 +9,9 @@ import AppLink from '../../presentational-components/shared/AppLink';
 import { getDateFormat } from '../../helpers/shared/helpers';
 import pathnames from '../../utilities/pathnames';
 import messages from '../../Messages';
+import { sortable } from '@patternfly/react-table';
+import { CheckIcon, CloseIcon } from '@patternfly/react-icons';
```
fhlavac, line 13: Can you combine this imports with the `OutlinedQuestionCircleIcon` import?

src/test/redux/reducers/group-reducer.test.js
fhlavac, line 1: This change is probably not needed

aferd: Yeah, I had a comment here on what test was failing and for me to look into the test, but no change was needed.

src/redux/reducers/group-reducer.js
fhlavac, line 1: @aferd  I am not really sure about these changes. In the follow-up ticket, we should try to make the state changes a bit more clear. Updating data for `selectedGroup` when you just expand a row on the groups page may cause confusion and unexpected errors in the future. I would be more for splitting the logic handling Group detail and Groups page - different actions and separated reducer logic to not change what is not necessary. Can you please take care of it with the pagination?

src/smart-components/group/group-table-helpers.js
```diff
@@ -9,6 +9,9 @@ import AppLink from '../../presentational-components/shared/AppLink';
...
+import { Table, TableBody, TableHeader, TableVariant } from '@patternfly/react-table';
```
fhlavac, line 14: Can you move "external" imports above the local ones?

fhlavac:
This definitely needs to be fixed
image

src/smart-components/group/group-table-helpers.js
fhlavac, line 13: We already have import from `@patternfly/react-icons` present, can you combine them?

fhlavac:
@aferd can you please remove the sort icons for now if the sorting does not work?
image

fhlavac:
we should take care of the loading state

fhlavac:
@aferd assuming the tests will be added as a part of the follow-up, right? Otherwise looks good :slightly_smiling_face:

src/smart-components/group/groups.js
```diff
@@ -176,10 +179,25 @@ const Groups = () => {
...
       : []),
   ];
+
   const data = groups.map((group) =>
     group.platform_default || group.admin_default ? { ...group, principalCount: `All${group.admin_default ? ' org admins' : ''}` } : group
   );
-  const rows = createRows(isAdmin, data, selectedRows);
+
+  const fetchExpandedRoles = (uuid) => dispatch(fetchRolesForGroup(uuid, defaultCompactSettings));
+  const fetchExpandedMembers = (uuid) => dispatch(fetchMembersForGroup(uuid, defaultCompactSettings));
```
fhlavac, line 188:
```suggestion
  const fetchExpandedRoles = (uuid) => useCallback(() => dispatch(fetchRolesForGroup(uuid, defaultCompactSettings, undefined, true)), [uuid]);
  const fetchExpandedMembers = (uuid) => useCallback(() => dispatch(fetchMembersForGroup(uuid, defaultCompactSettings, undefined, true)), [uuid]);

```

src/smart-components/group/groups.js
```diff
@@ -176,10 +179,25 @@ const Groups = () => {
...
+    if (!isOpen) {
+      setExpanded({ ...expanded, [rowData.uuid]: colIndex + Number(!isAdmin) });
+      colIndex + Number(!isAdmin) === 2 && fetchExpandedRoles(rowData.uuid);
+      colIndex + Number(!isAdmin) === 3 && fetchExpandedMembers(rowData.uuid);
+    } else {
+      setExpanded({ ...expanded, [rowData.uuid]: -1 });
+    }
+  };
+
+  const rows = createRows(isAdmin, data, selectedRows, expanded, isLoading);
```
fhlavac, line 200: this is probably not the isLoading flag we want - it's for `groups` data but we need to store another flag for group roles and members. Then it would be part of data (something like `group.roles.isLoading`) so no need to pass it separately

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,158 @@ DefaultPlatformPopover.propTypes = {
...
+                  )}
+                </Fragment>
+              </>
+            ) : (
+              name
+            ),
+          },
+          { title: roleCount, props: { isOpen: expanded[uuid] === 2 } },
+          {
+            title: principalCount,
```
fhlavac, line 86: a padding seems to be missing
```suggestion
            <div className={classNames({ 'pf-u-p-lg': isPlatformDefault || isAdminDefault })}>{principalCount}</div>,
```

src/smart-components/group/group-table-helpers.js
```diff
@@ -40,44 +42,158 @@ DefaultPlatformPopover.propTypes = {
...
+                  aria-label="Simple Table"
+                  variant={TableVariant.compact}
+                  cells={[
+                    { title: intl.formatMessage(messages.roleName) },
+                    { title: intl.formatMessage(messages.description) },
+                    { title: intl.formatMessage(messages.permissions) },
+                    { title: intl.formatMessage(messages.lastModified) },
+                  ]}
+                  rows={roles?.map((role) =>
+                    isLoading ? (
```
fhlavac, line 114: the isLoading condition should be probably above the `map` to work

src/redux/actions/group-actions.js
```diff
@@ -178,6 +178,7 @@ export const removeMembersFromGroup = (groupId, members) => {
...
   type: ActionTypes.FETCH_ROLES_FOR_GROUP,
   payload: GroupHelper.fetchRolesForGroup(groupId, false, pagination, options).catch(handleUuidError),
+  meta: { groupId },
```
fhlavac, line 181: you could pass something like `storeInGroups` boolean flag to switch between adding the data to `roles` or `selectedRole`. The same applies for members. It should also help you manage the loading boolean flag. Creating another actions for fetching expandable tables data is probably also correct if you want to go that way

aferd:
https://github.com/RedHatInsights/insights-rbac-ui/assets/86322239/81362492-0678-485f-b48e-d5040843fe05

karelhala (approved):

PR #1527, opened 2023-08-31
Add service accounts tab to Groups

fhlavac:
[RHCLOUD-28091)](https://issues.redhat.com/browse/RHCLOUD-28091)

- prepared layout for the service accounts tab in the Group detail
- added an "Add service account" button for the modal
- created an initial test case

image


src/smart-components/group/service-account/group-service-accounts.js
```diff
@@ -0,0 +1,195 @@
...
+import { fetchServiceAccountsForGroup } from '../../../redux/actions/group-actions';
+import { getBackRoute, getDateFormat } from '../../../helpers/shared/helpers';
+import PermissionsContext from '../../../utilities/permissions-context';
+import AppLink from '../../../presentational-components/shared/AppLink';
+import useAppNavigate from '../../../hooks/useAppNavigate';
+import messages from '../../../Messages';
+import pathnames from '../../../utilities/pathnames';
+import './group-service-accounts.scss';
+
+const createRows = (data, checkedRows = []) =>
```
Hyperkid123, line 21: Can we add a default value to the to data to avoid the `|| []` by the end of the reduce? I think its easier to ready.

fec.config.js
```diff
@@ -20,6 +20,6 @@ module.exports = {
       './IamUserAccess': path.resolve(__dirname, './src/entries/IamUserAccess.js'),
     },
     exclude: ['react-router-dom'],
-    shared: [{ 'react-router-dom': { singleton: true } }],
+    shared: [{ 'react-router-dom': { singleton: true } }, { '@unleash/proxy-client-react': { singleton: true, requiredVersion: '*' } }],
```
Hyperkid123, line 23: One more thing. The unleash should already be configured in the default webpack config. Can you verify it works without this configuration?

fhlavac: @Hyperkid123 it works, thank you!

Hyperkid123 (approved):

PR #1535, opened 2023-09-21
Add divider to Ungrouped systems

fhlavac:
[RHCLOUD-28241](https://issues.redhat.com/browse/RHCLOUD-28241)

image


src/smart-components/role/add-role/inventory-groups-role.js
```diff
@@ -157,13 +157,13 @@ const InventoryGroupsRole = (props) => {
...
       .includes(state[permissionID].filterValue.toLocaleLowerCase())
       ? [
-          {
-            name: 'null',
-            children: <FormattedMessage {...messages.ungroupedSystems} />,
-          },
+          <SelectOption key={`${permissionID}-${index + 1}`} value="null">
+            <FormattedMessage {...messages.ungroupedSystems} />
+          </SelectOption>,
+          <Divider component="li" key={5} />,
```
Hyperkid123, line 163: I am curious about these `key` choices. I think I understand why you made a change like this, just wondering if there are some better options to fill them. It just seems strange to see static values `5` and `+1`

fhlavac: just got distracted in the middle of rewriting the static value from the PF example, sorry :laughing:

Hyperkid123 (approved):

PR #1585, opened 2024-01-19
Enable Inventory groups editing

fhlavac:
[RHCLOUD-30558](https://issues.redhat.com/browse/RHCLOUD-30558)

Added a possibility to edit inventory groups in a similar way to cost-management resource definitions

image

image


src/smart-components/role/role-resource-definitions-table-helpers.js
```diff
@@ -1,5 +1,14 @@
...
-  data.reduce(
+import messages from '../../Messages';
+
+export const createRows = (data, isHosts, intl) => {
+  let finalData = data;
+  if (isHosts) {
+    finalData = data.filter((item) => item !== null);
+    if (finalData.length < data.length) {
+      finalData = [intl.formatMessage(messages.ungroupedSystems), ...finalData];
+    }
```
karelhala, line 9: This thing is used for checking if  `data` does not contain `null` value for ungroupped systems, right? I'd personally change it to something like
```
finalData = [...data.includes(null) ? [] : [intl.formatMessage(messages.ungroupedSystems)], ...finalData]
```

Even easier one could be
```
if (isHosts && !data.includes(null)) {
  finalData.unshift(intl.formatMessage(messages.ungroupedSystems))
}
```

src/smart-components/role/edit-resource-definitions-modal.js
```diff
@@ -2,32 +2,45 @@ import React, { useEffect, useReducer } from 'react';
...
-      ...acc,
-      ...value.map((r) => ({
-        value: r.value,
-        path: key,
-        label: r.value,
-      })),
-    ],
-    []
-  );
+const createOptions = (resources, isInventory, isHosts) =>
```
karelhala, line 21: We may want to adjust this later down the line as notifications will be added to this functionality.

karelhala (approved):
Looking good! Just two minor enhancements.

karelhala (approved):

PR #1615, opened 2024-04-11
RHCLOUD-31826: quick fix for sentry errors when switching between roles and groups

florkbr:
### Description

We have an issue where setting filters in the groups UI then switching to the roles UI is causing sentry errors. See attached RHCLOUD ticket and slack discussion here: https://redhat-internal.slack.com/archives/C05M0SNTLM8/p1712174912636909. This is causing several issues in our test suites. The issue is not reproducible every time in my observations, but if you set the filter and click back and forth enough it will eventually error.

I suspect this may have been introduced in https://github.com/RedHatInsights/insights-rbac-ui/pull/1601. After that change, while we load the roles UI - we make a request to load the adminGroup here: https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/smart-components/role/roles.js#L96. The adminGroup is then passed into our `createRows` helpers here: https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/smart-components/role/roles.js#L182 (which is throwing errors).

AFAIK there is no guarantee the adminGroup request will have loaded by the time the UI finishes loading and calls the `createRows` helper (hence adminGroup will be undefined until the request finishes and it makes its way into redux).

For now I've simply added an optional chaining check to the line that was throwing the sentry error. I think long term we need a few changes:
1. Do we really need to make the same admin group request every time we load the page? Or can it be cached?
2. If the admin group must be loaded before we do any table evaluation, we need to move the call to request it sooner in the page loading and stop the UI from rendering any further
3. Do we really need to use redux for this?
4. We _really_ should convert this to typescript - which likely would have alerted us to the fact that adminGroup can either be `undefined | Object`

https://issues.redhat.com/browse/RHCLOUD-31826

---

#### Before:
Screenshot 2024-04-11 at 6 11 13 PM

#### After:
Screenshot 2024-04-11 at 6 10 39 PM

---

### Checklist ☑️

##
##

### DIFF
--- a/src/smart-components/role/role-table-helpers.js
+++ b/src/smart-components/role/role-table-helpers.js
@@ -51,7 +51,7 @@ export const createRows = (data, selectedRows, intl, expanded, adminGroup) =>
                       group.description,
                       {
                         title:
-                          adminGroup.uuid === group.uuid ? null : (
+                          adminGroup?.uuid === group.uuid ? null : (
                             <AppLink
                               to={pathnames['roles-add-group-roles'].link.replace(':roleId', uuid).replace(':groupId', group.uuid)}
                               state={{ name: group.name }}


florkbr:
FWIW this change doesn't seem to break the functionality added in 1601: 
image

fhlavac (approved):
Thank you @florkbr for taking a look, regarding our points:

1-3) that definitely has to be changed. I should have realized that in that PR. In this case, it seems to me that it would be best to download and cache the Default admin access group once, because we only need its ID, which should not changed (I probably switched it in my thoughts with the Default access group, where the ID can be changed through customizations/reset).

4)I fully agree, we are already adding new files in TS and trying to move to TS when doing a larger refactoring somewhere. We should focus on this more during the tech debt sprint, or maybe a hackathon. There is a lot of work to be done

florkbr:
@fhlavac sounds good! I'll open a ticket and link it here for 1-3 and 4 separately.

florkbr:
Ticket for 1-3: https://issues.redhat.com/browse/RHCLOUD-32057

I'll take a look in the backlog if I can find one for 4. Otherwise I'll create one.

PR #1626, opened 2024-05-15
Update bundles.js

rabbott01:
I have added the following services to the to the list so their roles will appear in the MyUserAccess page.

- malware-detection
- ROS

I also removed the App and Data Services section as they no longer have products with permissions associated with RBAC.  

Page: https://console.redhat.com/iam/my-user-access?bundle=rhel
Associated Jira: [RHCLOUD-32064](https://issues.redhat.com/browse/RHCLOUD-32064)


fhlavac:
@rabbott01 useBundleApps.test.js needs to be updated as well for the CI to pass

PR #1635, opened 2024-07-17
Pull in rhecosystem changes (need FEO support)

florkbr:
### Description

description text...

[RHCLOUDXXXX](https://issues.redhat.com/browse/RHCLOUDXXXX)

---

#### After:

---

### Checklist ☑️
- [ ] PR only fixes one issue or story 
- [ ] Change reviewed for extraneous code 
- [ ] UI best practices adhered to 
- [ ] Commits squashed and meaningfully named 
- [ ] All PR checks pass locally (build, lint, test, E2E)

##
- [ ] _(Optional) QE: Needs QE attention (OUIA changed, perceived impact to tests, no test coverage)_
- [ ] _(Optional) QE: Has been mentioned_
- [ ] _(Optional) UX: Needs UX attention (end user UX modified, missing designs)_
- [ ] _(Optional) UX: Has been mentioned_
##

PR #1643, opened 2024-08-08
Add back token logic to invite users endpoints

florkbr:
### Description

We recently migrated the changes from RHEcosystem rbac-ui into this repo. With those changes I made some tweaks per code review comments. One of those tweaks was to remove the token logic (as we suspected chrome would auto apply the header for us): https://github.com/RedHatInsights/insights-rbac-ui/commit/d354f567c8caaed97c755935c069f79380f0aaa4#diff-195bc33b49d5867b3f0b7e71303727bad811474107d1ec281a1d14f8d855465dL36

However, as the target endpoints are on a different domain and not in our whitelist here: https://github.com/RedHatInsights/insights-chrome/blob/master/src/utils/iqeEnablement.ts#L17 - the auth header is not correctly applied.

For now I've added the logic we had back (although we should not be pulling the token from the global). I will open a tech debt issue to track improving this impl. Because these URLs are sensitive I have opted not to update our whitelist.

---

### Screenshots

NA

---

### Checklist ☑️

##
##


src/helpers/user/user-helper.js
```diff
@@ -55,9 +56,10 @@ function handleError(error, reject) {
 }
 
 export async function addUsers(usersData = { emails: [], isAdmin: undefined }) {
+  const token = await insights.chrome.auth.getToken();
```
florkbr, line 59: FWIW we should not be using the `insights` global. This was originally how the logic was migrated over from RHEcosystem in https://github.com/RedHatInsights/insights-rbac-ui/pull/1635. We can update the reducers/actions to accept a token from the useAuth hook in the future.

InsaneZein (approved):

catastrophe-brandon (approved):

florkbr:
Pipeline passing now due to https://github.com/RedHatInsights/insights-rbac-ui/pull/1652. Tested locally and everything is working as expected. I'm going to merge this to unblock inboundary and will keep an eye out for any issues.

PR #1655, opened 2024-08-29
Create Users & User Groups Page and Users List

CodyWMitchell:
### Create Users & User Groups Page and Users List

[RHCLOUD-34256](https://issues.redhat.com/browse/RHCLOUD-34256)
## Completed
- [x] Create a new page using DataView - do not reuse the old list with TableToolbarView
- [x] I can see all user information in the table (Org Admin Y/N, Username, Email Address, First Name, Last Name, Status and a list of associated User Groups (count or list).
- [x] I can choose how many users to view per page within the user list (5, 10, 20 [default view], 50, or 100)
- [x] Changes are hidden behind the platform.rbac.workspaces feature flag (Navigation to this page is only available with this feature flag)

## Will be handled by other tickets ([RHCLOUD-34816](https://issues.redhat.com/browse/RHCLOUD-34816))
- I can filter the results by org admin, username, email address, first name or last name
- I can sort the user results by any column in ascending or descending order. 
- Empty state is displayed if there are no results (when no rows are passed in. this will be handled on the DataView side)
- Loading state is displayed when the results are loading (when `isLoading` prop passed to DataView is true.)

---

### Screenshots

image
---

### Checklist ☑️

##
##

### DIFF
--- a/src/Routing.tsx
+++ b/src/Routing.tsx
@@ -39,7 +39,13 @@ const AddGroupServiceAccounts = lazy(() => import('./smart-components/group/serv
 const RemoveServiceAccountFromGroup = lazy(() => import('./smart-components/group/service-account/remove-group-service-accounts'));
 const QuickstartsTest = lazy(() => import('./smart-components/quickstarts/quickstarts-test'));
 
+const UsersAndUserGroups = lazy(() => import('./smart-components/access-management/users-and-user-groups'));
+
 const getRoutes = ({ enableServiceAccounts, isITLess }: Record<string, boolean>) => [
+  {
+    path: pathnames['users-and-user-groups'].path,
+    element: UsersAndUserGroups,
+  },
   {
     path: pathnames.overview.path,
     element: Overview,
--- a/src/smart-components/access-management/UsersTable.tsx
+++ b/src/smart-components/access-management/UsersTable.tsx
@@ -1,0 +1,125 @@
+import React, { useEffect, useCallback } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { useDataViewSelection, useDataViewPagination } from '@patternfly/react-data-view/dist/dynamic/Hooks';
+import { BulkSelect, BulkSelectValue } from '@patternfly/react-component-groups/dist/dynamic/BulkSelect';
+import { DataView } from '@patternfly/react-data-view/dist/dynamic/DataView';
+import { DataViewToolbar } from '@patternfly/react-data-view/dist/dynamic/DataViewToolbar';
+import { DataViewTable } from '@patternfly/react-data-view/dist/dynamic/DataViewTable';
+import { Pagination } from '@patternfly/react-core';
+import { ActionsColumn } from '@patternfly/react-table';
+import { fetchUsers } from '../../redux/actions/user-actions';
+import { mappedProps } from '../../helpers/shared/helpers';
+import { RBACStore } from '../../redux/store';
+import { UserProps } from '../user/user-table-helpers';
+import { useSearchParams } from 'react-router-dom';
+
+const COLUMNS: string[] = ['Username', 'Email', 'First name', 'Last name', 'Status', 'Org admin'];
+
+const ROW_ACTIONS = [
+  { title: 'Add to user group', onClick: () => console.log('ADD TO USER GROUP') },
+  { title: 'Remove from user group', onClick: () => console.log('REMOVE FROM USER GROUP') },
+];
+
+const PER_PAGE_OPTIONS = [
+  { title: '5', value: 5 },
+  { title: '10', value: 10 },
+  { title: '20', value: 20 },
+  { title: '50', value: 50 },
+  { title: '100', value: 100 },
+];
+
+const OUIA_ID = 'iam-users-table';
+
+const UsersTable: React.FunctionComponent = () => {
+  const dispatch = useDispatch();
+
+  const { users, totalCount } = useSelector((state: RBACStore) => ({
+    users: state.userReducer.users.data || [],
+    totalCount: state.userReducer.users.meta.count,
+  }));
+
+  const [searchParams, setSearchParams] = useSearchParams();
+  const pagination = useDataViewPagination({ perPage: 20, searchParams, setSearchParams });
+  const { page, perPage, onSetPage, onPerPageSelect } = pagination;
+
+  const selection = useDataViewSelection({ matchOption: (a, b) => a[0] === b[0] });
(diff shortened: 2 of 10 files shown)


src/Routing.tsx
```diff
@@ -210,6 +212,10 @@ const getRoutes = ({ enableServiceAccounts, isITLess }: Record<string, boolean>)
       },
     ],
   },
+  {
+    path: '/access-management/users-and-user-groups/*',
+    element: UsersAndUserGroups,
+  },
```
fhlavac, line 218:
```suggestion
  {
    path: '/users-and-user-groups/*',
    element: UsersAndUserGroups,
  },
```

this should help - you don't have to include the `access-management` section.

Re: "Before it was going to that page and not displaying my component there, but now it's automatically redirecting me from https://stage.foo.redhat.com:1337/iam/access-management/users-and-user-groups to https://stage.foo.redhat.com:1337/iam/user-access/users?status=Active&per_page=20&page=1 when the 'Users & User Groups' link is clicked."

This was because of a misconfiguration on the navigation side - I've opened a [PR](https://github.com/RedHatInsights/chrome-service-backend/pull/657) yesterday fixing that. Hope that helps

src/smart-components/access-management/UsersTable.js
```diff
@@ -0,0 +1,124 @@
...
+                  isSelected: isSelected(user),
+                }}
+              />
+              {Object.keys(columns).map((column) => (
+                <Td key={column}>{column === 'orgAdmin' ? (user[column] ? 'Yes' : 'No') : user[column]}</Td>
+              ))}
+            </Tr>
+          ))}
+        </Tbody>
+      </Table>
```
fhlavac, line 118: This can be replaced with `DataViewTable` https://react-data-view-pr-data-view-14.surge.sh/extensions/data-view/components

CodyWMitchell: I'm having some trouble making those rows selectable and adding the action menu on the right. Does the `DataViewTable` support that yet?

fhlavac: @CodyWMitchell here is a row selection example https://react-data-view-pr-data-view-14.surge.sh/extensions/data-view/functionality?page=1&perPage=5#selection-example

fhlavac: I've just added also an example for row actions - for now it works the same as with regular PF table https://react-data-view-pr-data-view-20.surge.sh/extensions/data-view/components#data-view-table

fhlavac:
The row selection seems to be broken

chrome-capture-2024-9-24

fhlavac:
@CodyWMitchell we should use typescript for new files

src/smart-components/access-management/UsersTable.js
```diff
@@ -0,0 +1,124 @@
...
+      } catch (err) {
+        setError('Failed to load users');
+        console.error(err);
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    loadUsers();
+  }, [page, perPage]);
```
fhlavac, line 60: It would be better to use Redux instead of directly fetching the data here - we can use similar logic to the former users page. Also, the promise middleware will help us catch the errors more gracefully.

src/smart-components/access-management/UsersTable.js
```diff
@@ -0,0 +1,124 @@
...
+
+const OUIA_ID = 'iam-users-table';
+
+const UsersTable = () => {
+  const [users, setUsers] = useState([]);
+  const [totalCount, setTotalCount] = useState(0);
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState(null);
+
+  const pagination = useDataViewPagination({ perPage: 20 });
```
fhlavac, line 34: Can you please enable pagination params sync with URL to keep the functionality we have in the Users page?

fhlavac:
Just small comments, overall looks good. Thanks, Cody!

src/smart-components/access-management/UsersTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+    loadUsers();
+  }, [page, perPage]);
+
+  const handleBulkSelect = (value: BulkSelectValue) => {
+    if (value === BulkSelectValue.none) {
+      onSelect(false);
+    } else if (value === BulkSelectValue.page) {
+      onSelect(true, users);
+    } else if (value === BulkSelectValue.all) {
+      onSelect(true);
```
fhlavac, line 78:
```suggestion
    } else if (value === BulkSelectValue.nonePage) {
      onSelect(false, rows);
```

This should fix the select - the all option is not needed with the pagination

src/smart-components/access-management/UsersTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+    />
+  );
+
+  return (
+    <DataView ouiaId={OUIA_ID} selection={selection}>
+      <DataViewToolbar
+        ouiaId={`${OUIA_ID}-header-toolbar`}
+        bulkSelect={
+          <BulkSelect
+            isDataPaginated={true}
```
fhlavac, line 117:
```suggestion
            isDataPaginated
```

src/smart-components/access-management/UsersTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { DataView } from '@patternfly/react-data-view/dist/dynamic/DataView';
+import { DataViewToolbar } from '@patternfly/react-data-view/dist/dynamic/DataViewToolbar';
+import { DataViewTable } from '@patternfly/react-data-view/dist/dynamic/DataViewTable';
+import { Pagination } from '@patternfly/react-core';
+import { ActionsColumn } from '@patternfly/react-table';
+import { fetchUsers } from '../../redux/actions/user-actions';
+import { mappedProps } from '../../helpers/shared/helpers';
+import { RBACStore } from '../../redux/store';
+
+interface User {
```
fhlavac, line 14: It's just nitpicking, but not sure if here is the best place to define this interface - it may be reused in multiple places. How about moving it to the user-reducer or similar place?

fhlavac:
@CodyWMitchell looks great! just the pagination params sync with URL and we are golden 🏅

src/smart-components/access-management/UsersTable.tsx
CodyWMitchell: Updated to use the existing interface in `/user/user-table-helpers.tsx`

fhlavac (approved):
Great work! Thanks @CodyWMitchell 🎉

PR #1696, opened 2024-11-06
feat(rolestable): delete roles modal

InsaneZein:
### Description
This allows user to delete Roles. Currently deleting roles api doesn't exist so in the meantime, clicking "delete role" button will just spit out a console.log message. 

[RHCLOUD-32241](https://issues.redhat.com/browse/RHCLOUD-32241)

---

### Checklist ☑️
- [ ] PR only fixes one issue or story 
- [ ] Change reviewed for extraneous code 
- [ ] Commits squashed and meaningfully named 
- [ ] All PR checks pass locally (build, lint, test, E2E)

##
##

### DIFF
--- a/src/smart-components/role/RolesTable.tsx
+++ b/src/smart-components/role/RolesTable.tsx
@@ -9,8 +9,8 @@ import { DataViewEventsProvider, EventTypes, useDataViewEventsContext } from '@p
 import { ButtonVariant, Drawer, DrawerContent, DrawerContentBody, PageSection, Pagination } from '@patternfly/react-core';
 import { ActionsColumn } from '@patternfly/react-table';
 import ContentHeader from '@patternfly/react-component-groups/dist/esm/ContentHeader';
-import { fetchRolesWithPolicies } from '../../redux/actions/role-actions';
-import { useIntl } from 'react-intl';
+import { fetchRolesWithPolicies, removeRole } from '../../redux/actions/role-actions';
+import { FormattedMessage, useIntl } from 'react-intl';
 import messages from '../../Messages';
 import { mappedProps } from '../../helpers/shared/helpers';
 import { Role } from '../../redux/reducers/role-reducer';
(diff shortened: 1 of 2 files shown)


InsaneZein:
I realized that sometimes selectedRole is undefined so the modal shows undefined instead of the role name.

src/smart-components/role/RolesTable.tsx
```diff
@@ -118,6 +135,23 @@ const RolesTable: React.FunctionComponent<RolesTableProps> = ({ selectedRole = u
     <React.Fragment>
       <ContentHeader title="Roles" subtitle={''} />
       <PageSection isWidthLimited>
+        {isDeleteModalOpen && (
+          <WarningModal
+            ouiaId={`${ouiaId}-remove-role-modal`}
+            isOpen={isDeleteModalOpen}
```
karelhala, line 141: Based on the mocks the modal is missing checkbox
```suggestion
            isOpen={isDeleteModalOpen}
            withCheckbox
```

karelhala (changes requested):
Looks like some of the messages needs to be changed to role's messages. Also when I pulled the PR I found it's missing the toolbar actions which should allow to remove multiple roles.

karelhala:
@InsaneZein I've checked the API and it has a delete role option https://console.redhat.com/docs/api/rbac/v1#operations-Role-deleteRole it should however be available only on custom roles. If the role is custom can be checked in role definition if `system = false`

karelhala (changes requested):
The modal is using groups messages and looks like this:
Screenshot 2024-11-13 at 9 17 04

But the messages should be:
@
image-2024-09-09-12-24-18-530


The DataView is missing an action button to remove multiple roles. Also selecting single item selects the whole table.

InsaneZein:
Ah ok the task made it sound like this was just for deleting a singular Role. I can work on adding an action button to delete multiple roles in another PR.

karelhala (approved):

PR #1732, opened 2024-12-13
feat: create edit user group page

CodyWMitchell:
### Description

Created the Edit User Group page

- [x] I can edit an existing user group by performing any or all of the following actions: 
- [x] I can rename the user group.
- [x] I can update the description of the user group.
- [x] I am notified if the updated user group name has already been used for my organization.
- [x] I can confirm or cancel the request.
- [x] Upon cancellation, I am redirected to the previous page/task. 
- [x] Upon saving, I am presented with a success message and redirected to the User Group list. 
- [x] I am presented with an error message if the user group is not saved successfully.
### Will be handled in future ticket
- [ ] I can add or remove principal users and service accounts from the user group from their respective tabs.

Currently values are diffed and displayed, then logged to the console. Another ticket will handle sending out the appropriate requests.

[RHCLOUD-32236](https://issues.redhat.com/browse/RHCLOUD-32236)

Improve the integration between DataView and data-driven-forms, and potentially create a custom DataView component which could simplify this logic and enable disabling the submit button when no changes have been made here:
[RHCLOUD-37436](https://issues.redhat.com/browse/RHCLOUD-37436) & [RHCLOUD-37435](https://issues.redhat.com/browse/RHCLOUD-37435)

---

#### After:
image

---

### Checklist ☑️
- [X] PR only fixes one issue or story 
- [X] Change reviewed for extraneous code 
- [X] UI best practices adhered to 
- [X] Commits squashed and meaningfully named 
- [X] All PR checks pass locally (build, lint, test, E2E)

##
##

### DIFF
--- a/src/Routing.tsx
+++ b/src/Routing.tsx
@@ -47,6 +47,7 @@ const QuickstartsTest = lazy(() => import('./smart-components/quickstarts/quicks
 const UsersAndUserGroups = lazy(() => import('./smart-components/access-management/users-and-user-groups/users-and-user-groups'));
 const UsersView = lazy(() => import('./smart-components/access-management/users-and-user-groups/users/UsersView'));
 const UserGroupsView = lazy(() => import('./smart-components/access-management/users-and-user-groups/user-groups/UserGroupsView'));
+const EditUserGroup = lazy(() => import('./smart-components/access-management/users-and-user-groups/user-groups/edit-user-group/EditUserGroup'));
 
 const getRoutes = ({ enableServiceAccounts, isITLess, isWorkspacesFlag, isCommonAuthModel }: Record<string, boolean>) => [
   {
@@ -74,6 +75,10 @@ const getRoutes = ({ enableServiceAccounts, isITLess, isWorkspacesFlag, isCommon
         ],
       },
     ],
+  },
+  {
+    path: pathnames['users-and-user-groups-edit-group'].path,
+    element: EditUserGroup,
   },
   {
     path: pathnames.overview.path,
--- a/cypress/e2e/users-and-user-groups.cy.ts
+++ b/cypress/e2e/users-and-user-groups.cy.ts
@@ -171,4 +171,12 @@ describe('Users and User Groups page', () => {
     cy.get('[data-ouia-component-id="iam-user-groups-table-actions-dropdown-action-0"]').click();
     cy.get('[data-ouia-component-id="add-group-wizard"]').should('exist');
   });
+
+  it('should be able to open Edit User Group page from row actions', () => {
+    cy.get('[data-ouia-component-id="user-groups-tab-button"]').click();
+    cy.get('[data-ouia-component-id^="iam-user-groups-table-table-td-0-7"]').click();
+    cy.get('[data-ouia-component-id^="iam-user-groups-table-table-td-0-7"] button').contains('Edit user group').click();
+    cy.url().should('include', '/iam/access-management/users-and-user-groups/edit-group');
+    cy.get('[data-ouia-component-id="edit-user-group-form"]').should('be.visible');
+  });
 });
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -2528,4 +2528,9 @@ export default defineMessages({
     description: 'create user group button label',
     defaultMessage: 'Create user group',
   },
+  selectUsersAndOrServiceAccounts: {
+    id: 'selectUsersAndOrServiceAccounts',
+    description: 'select users and/or service accounts label',
+    defaultMessage: 'Select users and/or service accounts',
+  },
 });
--- a/src/redux/reducers/group-reducer.ts
+++ b/src/redux/reducers/group-reducer.ts
@@ -36,7 +36,7 @@ export interface GroupStore {
     filters: any;
     pagination: { count: number };
   };
-  selectedGroup: {
+  selectedGroup: Group & {
     addRoles: any;
     members: { meta: PaginationDefaultI; data?: any[] };
     serviceAccounts: { meta: PaginationDefaultI; data?: any[] };
--- a/src/smart-components/access-management/users-and-user-groups/user-groups/UserGroupsTable.tsx
+++ b/src/smart-components/access-management/users-and-user-groups/user-groups/UserGroupsTable.tsx
@@ -177,7 +177,7 @@ const UserGroupsTable: React.FunctionComponent<UserGroupsTableProps> = ({
               items={[
                 {
                   title: intl.formatMessage(messages['usersAndUserGroupsEditUserGroup']),
-                  onClick: () => console.log('EDIT USER GROUP'),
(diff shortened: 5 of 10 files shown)


fhlavac:
we should introduce some loading state to prevent this from happening
chrome-capture-2025-1-7

src/Routing.tsx
```diff
@@ -75,6 +76,10 @@ const getRoutes = ({ enableServiceAccounts, isITLess, isWorkspacesFlag, isCommon
       },
     ],
   },
+  {
```
fhlavac, line 79: should be a child route

src/smart-components/access-management/EditUserGroupUsersAndServiceAccounts.tsx
```diff
@@ -0,0 +1,50 @@
...
+    });
+  };
+
+  const handleTabSelect = (_: React.MouseEvent<HTMLElement, MouseEvent>, key: string | number) => {
+    setActiveTabKey(Number(key));
+  };
+
+  return (
+    <React.Fragment>
+      <FormGroup label="Select users and/or service accounts">
```
fhlavac, line 38: Should be localized

src/smart-components/access-management/EditUserGroupUsers.tsx
```diff
@@ -0,0 +1,131 @@
...
+            pageCount={users.length}
+            selectedCount={selected.length}
+            totalCount={totalCount}
+            pageSelected={pageSelected}
+            pagePartiallySelected={pagePartiallySelected}
+            onSelect={handleBulkSelect}
+          />
+        }
+      />
+      <DataViewTable variant="compact" columns={['Org Admin', 'Username', 'Email', 'First name', 'Last name', 'Status']} rows={rows} />
```
fhlavac, line 126: Should be localized

fhlavac:
Compared to mockups we should have local breadcrumbs and the "Name" field should be required
image

image

fhlavac:
It would be also nice to disable "Submit" button when there are no changes - it should be pretty easy using the form state

src/smart-components/access-management/EditUserGroupUsers.tsx
fhlavac, line 1: Maybe we could just reuse the existing`UsersTable`, what do you think? Also, it might be good to add loading state when changing pages (the `UsersTable` already has them)

src/smart-components/access-management/EditUserGroupServiceAccounts.tsx
```diff
@@ -0,0 +1,178 @@
...
+    };
+  }, []);
+
+  const { serviceAccounts, status } = useSelector(reducer);
+  const calculateTotalCount = () => {
+    if (!serviceAccounts) return 0;
+    const currentCount = (page - 1) * perPage + serviceAccounts.length;
+    return status === LAST_PAGE ? currentCount : currentCount + 1;
+  };
+  const totalCount = calculateTotalCount();
```
fhlavac, line 62: could be memoized

fhlavac:
@CodyWMitchell looks good! Just few comments

src/smart-components/access-management/EditUserGroupUsers.tsx
CodyWMitchell: Since selection is being used for enabling and disabling users here - with the mocks showing the `Org Admin` column shuffled around and no action buttons - I made this into its own component instead of reusing the users table. I can try to make the original users table work for both cases if you think that's better though. Either way, I'll make sure to add loading states to this version. :+1:

src/smart-components/access-management/EditUserGroup.tsx
```diff
@@ -0,0 +1,114 @@
...
+      if (serviceAccounts.removed.length > 0) {
+        console.log(`Service accounts removed: ${serviceAccounts.removed}`);
+      }
+      returnToPreviousPage();
+    }
+  };
+
+  return (
+    <React.Fragment>
+      <ContentHeader title={intl.formatMessage(Messages.usersAndUserGroupsEditUserGroup)} subtitle={''} />
```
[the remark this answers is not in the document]
CodyWMitchell: It looks like typescript requires the `subtitle` prop in the `ContentHeader`, it's not marked as optional. I see an error when I remove it.

src/smart-components/access-management/EditUserGroupUsers.tsx
fhlavac: @CodyWMitchell columns shuffled around is an interesting thing --it might be good to ask Margot, who prepared the mockups if that's intentional. But, I agree that this table will probably be much simpler than the regular Users view - it's probably a safer solution in this case 👍

src/smart-components/access-management/EditUserGroup.tsx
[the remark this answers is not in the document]
fhlavac: Ahh, I see. I'll open an issue in the component groups

src/Routing.tsx
CodyWMitchell: image
Rendering this as a child route adds the **Users and Users Groups** header above - is there a way to avoid this?

src/smart-components/access-management/EditUserGroupServiceAccounts.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import { fetchServiceAccountsForGroup } from '../../redux/actions/group-actions';
+import { RBACStore } from '../../redux/store';
+import { mappedProps } from '../../helpers/shared/helpers';
+import { fetchServiceAccounts } from '../../redux/actions/service-account-actions';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
+import { ServiceAccountsState } from '../../redux/reducers/service-account-reducer';
+import { LAST_PAGE, ServiceAccount } from '../../helpers/service-account/service-account-helper';
+import { BulkSelect, BulkSelectValue } from '@patternfly/react-component-groups';
+import DateFormat from '@redhat-cloud-services/frontend-components/DateFormat';
```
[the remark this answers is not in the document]
CodyWMitchell: The mocks show the time formatted as `1 month ago`. I didn't see a straightforward way to do this with the [PatternFly Timestamp](https://www.patternfly.org/components/timestamp/#custom-format) in the docs without defining it custom somehow.
image

fhlavac:
@CodyWMitchell when opening the detail, the URL should contain `access-management`, not `user-access`
image

fhlavac:
The submit button looks to be enabled even if the form is pristine 
image

fhlavac:
chrome-capture-2025-1-24
After submitting the form, the table does not seem to be updated.

karelhala:
@fhlavac for the breadcrumbs the workspaces effort has them broken and we'll have to fix a few of the screens. For the pristine state, that's connected to DDF and DW interaction and we'll have to write custom component and fix it in tech debt. As for the non updated values @CodyWMitchell will create a ticket to address this in a different ticket since this looks like API race condition.

PR #1749, opened 2025-01-23
Add filtering for workspaces

apinkert:
For [RHCLOUD-35021](https://issues.redhat.com/browse/RHCLOUD-35021). After speaking with @karelhala, we decided the other filters were not necessary for this PR (creator hasn't been implemented yet and parent search was redundant). 

From our conversation:
 - A filtered workspace should show its parents
 - A filtered workspace should not show its children

https://github.com/user-attachments/assets/6c17c712-1f86-41c6-94f4-1eb5c75c14b5

### DIFF
--- a/src/smart-components/workspaces/WorkspaceListTable.tsx
+++ b/src/smart-components/workspaces/WorkspaceListTable.tsx
@@ -20,7 +20,6 @@ import AppLink from '../../presentational-components/shared/AppLink';
 import pathnames from '../../utilities/pathnames';
 import messages from '../../Messages';
 import useAppNavigate from '../../hooks/useAppNavigate';
-import DataViewFilters from '@patternfly/react-data-view/dist/cjs/DataViewFilters';
 
 interface WorkspaceFilters {
   name: string;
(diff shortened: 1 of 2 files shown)


src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -83,6 +130,12 @@ const WorkspaceListTable = () => {
               onSelect={handleBulkSelect}
             />
           }
+          clearAllFilters={clearAllFilters}
+          filters={
+            <DataViewFilters onChange={(_e, values) => onSetFilters(values)} values={filters}>
```
InsaneZein, line 135: I don't think `DataViewFilters` is necessary since we just have the single filter. Just using `DataViewTextFilter` is fine.

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -83,6 +130,12 @@ const WorkspaceListTable = () => {
...
+              <DataViewTextFilter filterId="name" title="Name" placeholder="Filter by name" />
```
InsaneZein, line 136: We should add an `ouiaId` here

src/smart-components/workspaces/WorkspaceListTable.tsx
apinkert: Will remove. This was a remnant of needing multiple filters.

src/smart-components/workspaces/WorkspaceListTable.tsx
apinkert: Added

InsaneZein (approved):

PR #1781, opened 2025-03-20
Add asset card for insights and openshift for workspaces

apinkert:
For [RHCLOUD-35614](https://issues.redhat.com/browse/RHCLOUD-35614)

image

### DIFF
--- a/src/smart-components/workspaces/workspace-detail/AssetsCards.tsx
+++ b/src/smart-components/workspaces/workspace-detail/AssetsCards.tsx
@@ -32,20 +32,17 @@ const AssetsCards: React.FunctionComponent<AssetsCardsProps> = ({ workspaceName 
     lg: '150px',
     xl: '200px',
     '2xl': '300px',
-  }
-  const AssetsCardsIconWidths = { 
+  };
+  const AssetsCardsIconWidths = {
     default: '48px',
-  }
+  };
 
   return (
     <Panel>
       <PanelHeader>{intl.formatMessage(Messages.assetManagementOverview)}</PanelHeader>
       <PanelMain>
         <PanelMainBody>
-          <Gallery
-            hasGutter
-            minWidths={AssetsCardsWidths}
-          >
+          <Gallery hasGutter minWidths={AssetsCardsWidths}>
             <Card>
               <CardHeader>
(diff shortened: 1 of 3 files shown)


apinkert:
@karelhala I'm still a bit unfamiliar with what assets are going to available for workspaces and the perms needed to show them. Is simply directing them to `/insights/inventory?workspace=[workspaceName]` and `/openshift/?workspace=[workspaceName]` enough?

src/smart-components/workspaces/workspace-detail/AssetsCards.tsx
```diff
@@ -1,9 +1,82 @@
...
+              </CardHeader>
+              <CardTitle>{intl.formatMessage(Messages.assetManagementInsights)}</CardTitle>
+              <CardBody>{intl.formatMessage(Messages.assetManagementInsightsOverview)}</CardBody>
+              <CardFooter>
+                <a href={InsightsNavURL}>
+                  {intl.formatMessage(Messages.assetManagementInsightsNav)}
+                  <Icon className="pf-v5-u-ml-sm" isInline>
+                    <ArrowRightIcon />
+                  </Icon>
+                </a>
```
karelhala, line 57: We should probably wrap these links with Button with component set to `a` and onClick to use navigate function. This way the page would not force refreshed and only navigated to the proper screen.

src/smart-components/workspaces/workspace-detail/AssetsCards.tsx
```diff
@@ -1,9 +1,82 @@
...
+              </CardFooter>
+            </Card>
+            <Card>
+              <CardHeader>
+                <Brand src={OpenShiftIcon} alt="OpenShift logo" style={{ width: '48px' }} />
```
karelhala, line 62: We should not inject styles @epwinchell do you have any suggestion for Brand component to have a specific width?

karelhala (commented):
Looking good, just two minor adjustments.

src/smart-components/workspaces/workspace-detail/AssetsCards.tsx
apinkert: @karelhala updated `Brand` to use its `widths` option instead

karelhala (approved):

PR #1794, opened 2025-04-10
feat(workspacewizard): expose workspace wizard through fec

InsaneZein:
### Description
part of [RHCLOUD-34248](https://issues.redhat.com/browse/RHCLOUD-34248)

I need to expose the Workspaces Wizard so that I can use it in the widget-layout repo

### Checklist ☑️
- [ ] PR only fixes one issue or story 
- [ ] Change reviewed for extraneous code 
- [ ] Commits squashed and meaningfully named 
- [ ] All PR checks pass locally (build, lint, test, E2E)

##
##

### DIFF
--- a/src/smart-components/workspaces/create-workspace/CreateWorkspaceWizard.tsx
+++ b/src/smart-components/workspaces/create-workspace/CreateWorkspaceWizard.tsx
@@ -10,14 +10,8 @@ import { createWorkspace } from '../../../redux/actions/workspaces-actions';
 import SetEarMark from './SetEarMark';
 import Review from './Review';
 import SetDetails from './SetDetails';
-import { Provider } from 'react-redux';
-import registry, { RegistryContext } from '../../../utilities/store';
-import { IntlProvider } from 'react-intl';
-import messages from '../../../locales/data.json';
 
-export const locale = 'en';
-
-interface CreateWorkspaceWizardProps {
+export interface CreateWorkspaceWizardProps {
   afterSubmit: () => void;
   onCancel: () => void;
 }
@@ -58,20 +52,4 @@ export const CreateWorkspaceWizard: React.FunctionComponent<CreateWorkspaceWizar
   );
 };
 
-const CreateWorkspaceWizardWrapper: React.FC<CreateWorkspaceWizardProps> = ({ ...props }) => {
-  return (
-    <IntlProvider locale={locale} messages={messages[locale]}>
-      <RegistryContext.Provider
-        value={{
-          getRegistry: () => registry,
-        }}
-      >
-        <Provider store={registry.getStore()}>
-          <CreateWorkspaceWizard {...props} />
-        </Provider>
-      </RegistryContext.Provider>
-    </IntlProvider>
-  );
-};
-
-export default CreateWorkspaceWizardWrapper;
+export default CreateWorkspaceWizard;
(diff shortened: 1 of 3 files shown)


src/smart-components/workspaces/create-workspace/CreateWorkspaceWizard.tsx
```diff
@@ -52,4 +54,14 @@ export const CreateWorkspaceWizard: React.FunctionComponent<CreateWorkspaceWizar
...
 };
 
-export default CreateWorkspaceWizard;
+const CreateWorkspaceWizardWrapper: React.FC<{ store?: Store } & CreateWorkspaceWizardProps> = ({ store, ...props }) => {
+  return store ? (
+    <Provider store={store}>
+      <CreateWorkspaceWizard {...props} />
+    </Provider>
+  ) : (
+    <CreateWorkspaceWizard {...props} />
```
Hyperkid123, line 63: If the store is not prpvided, should we create a default store or something? Redux is not configured as a singleton and the provider might not get shared in the tree and any references to redux will fail at runtime.

InsaneZein: I'm not sure. I was following the examples of the other components I saw exposed through fec.

Hyperkid123: This is not really about FEC or module federation. If you do not provide a redux store, once you call `useDispatch`, for example, there will be no context from which the hook can access the store.

karelhala: We have two use cases for this component:
* in RBAC - the store is to be expected from the RBAC application
* outside of RBAC - the store needs to be passed down to the component when imported as module

@Hyperkid123 do you have any recomendations on this two usecases? In already used components we've followed this path of checking for store in props. If it was provided it means it's from outside world and hence we have to wrap it in provider. The reason why we need outside store is to properly bootstrap the store in the parent component, reducers and middlewares. But since this is fairly isolated component we could create a new component which is exposed using the module federation with custom store and the store can be bootstrapped similat to https://github.com/RedHatInsights/insights-rbac-ui/blob/master/src/AppEntry.js#L17

Hyperkid123: While using this in RBAC, you will directly use the `CreateWorkspaceWizard ` which "expects the provider to be defined by the RBAC app."

This newly exposed component is a new version of "RBAC application," if you will. It has to provide all the necessary context the `CreateWorkspaceWizard ` expects. One of which is the redux store provider. I think you will also require the `IntlProvider` (If my memory serves me right, RBAC uses that).

The question is not necessary about the instance of the store (and other context values), but the existence of it.  Unless you explicitly need some initial state, I'd not pass the store as a prop. And even if you do, It would be better to pass the initial state and create the store internally based on that.

These exposed modules should be as isolated as possible. If some initialization is needed, create a well defined api. But do not pass the context around. That can cause dependency requirement on the consumer and that can breakdown the module independence.

src/smart-components/workspaces/create-workspace/CreateWorkspaceWizard.tsx
```diff
@@ -52,4 +58,20 @@ export const CreateWorkspaceWizard: React.FunctionComponent<CreateWorkspaceWizar
...
+      >
+        <Provider store={registry.getStore()}>
+          <CreateWorkspaceWizard {...props} />
+        </Provider>
+      </RegistryContext.Provider>
+    </IntlProvider>
+  );
+};
+
+export default CreateWorkspaceWizardWrapper;
```
karelhala, line 77: Let's move this into seperate file `CreateWorkspaceWizardModule` this way we don't have to change previous exposed component and have full control over what is being used outside RBAC.

karelhala (approved):

PR #1838, opened 2025-05-21
Update edit and create validators with workspace IDs to disallow duplicates 

apinkert:
For [RHCLOUD-39757](https://issues.redhat.com/browse/RHCLOUD-39757). On edit, the workspace name validator would always come up as a duplicate because its name already existed in the `allWorkspaces` array which would prevent being able to submit on description change only.

### DIFF
--- a/src/smart-components/workspaces/EditWorkspaceModal.tsx
+++ b/src/smart-components/workspaces/EditWorkspaceModal.tsx
@@ -56,11 +56,13 @@ const EditWorkspaceModal: React.FunctionComponent<EditWorkspaceModalProps> = ({ 
           component: componentTypes.TEXT_FIELD,
           validate: [
             { type: validatorTypes.REQUIRED },
-            (value: string) => {
+            (value: string, { id }: any) => {
               if (value === initialFormData?.name) {
                 return undefined;
               }
-              const isDuplicate = allWorkspaces.some((existingWorkspace) => existingWorkspace.name.toLowerCase() === value?.toLowerCase());
+              const isDuplicate = allWorkspaces.some(
+                (existingWorkspace) => existingWorkspace.name.toLowerCase() === value?.toLowerCase() && existingWorkspace.id !== id
+              );
               return isDuplicate ? intl.formatMessage(messages.groupNameTakenTitle) : undefined;
             },
           ],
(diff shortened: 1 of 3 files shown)


src/redux/reducers/workspaces-reducer.ts
```diff
@@ -44,6 +44,10 @@ const setWorkspace = (state: WorkspacesStore, { payload }: { payload: { data: Wo
   isLoading: false,
 });
 
+export const isWorkspace = (data: unknown): data is Workspace => {
+  return typeof data === 'object' && data !== null && 'id' in data && 'parent_id' in data && 'type' in data;
```
Hyperkid123, line 48: Should we also check what type are the object properties? For example, is this a valid workspaces:
```js
const ws = {
  id: null,
  parent_id: null,
  type: undefined
}

'type' in ws // yields true even if the attribute is undefined or null
```

src/smart-components/workspaces/EditWorkspaceModal.tsx
```diff
@@ -56,12 +57,17 @@ const EditWorkspaceModal: React.FunctionComponent<EditWorkspaceModalProps> = ({
           component: componentTypes.TEXT_FIELD,
           validate: [
             { type: validatorTypes.REQUIRED },
-            (value: string) => {
-              if (value === initialFormData?.name) {
-                return undefined;
+            (value: string, currData: unknown | undefined) => {
+              if (isWorkspace(currData)) {
```
Hyperkid123, line 61: We can make the `isWorkspace` a TS type guard. It will remove the necessity for typecasting in the branch.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

src/smart-components/workspaces/create-workspace/schema.tsx
```diff
@@ -104,6 +107,15 @@ export const schemaBuilder = (enableBillingFeatures: boolean) => {
                   {
                     type: validatorTypes.REQUIRED,
                   },
+                  (value: string, currData: unknown | undefined) => {
+                    if (isWorkspace(currData)) {
```
Hyperkid123, line 111: ^^

src/redux/reducers/workspaces-reducer.ts
```diff
@@ -44,6 +44,10 @@ const setWorkspace = (state: WorkspacesStore, { payload }: { payload: { data: Wo
   isLoading: false,
 });
 
+export const isWorkspace = (data: unknown | Workspace): data is Workspace => {
```
apinkert, line 47: @Hyperkid123 Great suggestion. This is definitely cleaner and allowed me to delete the casting in each validator for `id`

karelhala (approved):

PR #1849, opened 2025-06-04
Revert "fix(bug): fix circular structure error in user activation toggle"

InsaneZein:
Reverts RedHatInsights/insights-rbac-ui#1847


shannon-donahue (approved):
LGTM

PR #1850, opened 2025-06-04
Add conditional edit/delete button disabling

apinkert:
For [RHCLOUD-39840](https://issues.redhat.com/browse/RHCLOUD-39840)

Conditionally disables a workspace's edit/delete button based on workspace type as well as user permissions. Also adds basic e2e tests to test workspace type conditions.

image

## Summary by Sourcery

Add conditional disabling of workspace edit and delete actions based on workspace type and user RBAC permissions.

New Features:
- Fetch and store user RBAC permissions via Chrome to control workspace actions

Enhancements:
- Extend workspace type enum to include 'ungrouped-hosts'
- Disable edit/delete buttons dynamically according to workspace type and user permission

Tests:
- Add Cypress e2e tests to verify edit/delete button states for root, default, and standard workspaces based on permissions

### DIFF
--- a/src/smart-components/workspaces/WorkspaceListTable.tsx
+++ b/src/smart-components/workspaces/WorkspaceListTable.tsx
@@ -38,10 +38,46 @@ interface WorkspaceFilters {
   name: string;
 }
 
-interface PermissionObject {
-  permission: string;
-  resourceDefinitions: string[];
+interface PermissionResource {
+  key: string;
+  value: string[];
+  operation: string;
 }
+
+interface PermissionFilter {
+  attributeFilter: PermissionResource;
+}
+
+interface Permission {
+  permission: string;
+  resourceDefinitions: PermissionFilter[];
+}
+
+const isValidEditType = (workspace: Workspace) => {
+  switch (workspace.type) {
+    case 'root':
+      return false;
+    case 'default':
+      return true;
+    case 'ungrouped-hosts':
+      return false;
+    case 'standard':
+      return true;
+  }
+};
+
+const isValidDeleteType = (workspace: Workspace) => {
+  switch (workspace.type) {
+    case 'root':
+      return false;
+    case 'default':
+      return false;
+    case 'ungrouped-hosts':
+      return false;
+    case 'standard':
+      return true;
+  }
+};
 
 const mapWorkspacesToHierarchy = (workspaceData: Workspace[]): Workspace | undefined => {
   const idMap = new Map();
@@ -105,7 +141,7 @@ const WorkspaceListTable = () => {
   const chrome = useChrome();
   const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
   const [currentWorkspaces, setCurrentWorkspaces] = useState<Workspace[]>([]);
-  const [userPermissions, setUserPermissions] = useState<PermissionObject>({
+  const [userPermissions, setUserPermissions] = useState<Permission>({
     permission: '',
     resourceDefinitions: [],
   });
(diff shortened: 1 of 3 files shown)


src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -143,14 +192,18 @@ const WorkspaceListTable = () => {
...
+                  isSeparator: true,
                 },
-                { title: <Divider component="li" key="divider" /> },
                 {
                   title: 'Delete workspace',
                   onClick: () => {
                     handleModalToggle([workspace]);
                   },
-                  isDisabled: workspace.children && workspace.children.length > 0,
+                  isDisabled: workspace.children && workspace.children.length > 0 && !canModify(workspace, 'delete'),
```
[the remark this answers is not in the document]
karelhala: I think this condition is not correct, we want to disable the button if the WS has any children OR if user can't modify it.
```suggestion
                  isDisabled: workspace.children && workspace.children.length > 0 || !canModify(workspace, 'delete'),
```

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -107,6 +118,47 @@ const WorkspaceListTable = () => {
...
+      case 'root':
+        return false;
+      case 'default':
+        return false;
+      case 'ungrouped-hosts':
+        return false;
+      case 'standard':
+        return true;
+    }
+  };
```
karelhala, line 145: Let's move these two functions out of the component so we don't have to re-run their calculataion on each render.

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -107,6 +118,47 @@ const WorkspaceListTable = () => {
...
+
+  const canModify = (workspace: Workspace, action: 'edit' | 'delete') => {
+    if (
+      userPermissions.permission === 'inventory:groups:write' &&
```
karelhala, line 149: There is also `inventory:groups:*` permission which allows write. Let's change this
```suggestion
      ['inventory:groups:write', 'inventory:groups:*'].includes(userPermissions.permission) &&
```

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -107,6 +118,47 @@ const WorkspaceListTable = () => {
...
+      (userPermissions.resourceDefinitions.length === 0 || userPermissions.resourceDefinitions.includes(workspace.id))
```
karelhala, line 150: The full shape of Resouce definition is as:
```
[{
    "resourceDefinitions":[
      {"attributeFilter": {"key":"group.id", "value": string[], "operation":"in"} }
    ],
    "permission":"inventory:groups:write"
}]
```

We are looking for the `value`
```suggestion
      (userPermissions.resourceDefinitions.length === 0 || userPermissions.resourceDefinitions.some((item) => item.attributeFilter.value.includes(workspace.id))
```

We could add a bit more safe guards here since the operation can be different and key as well. But that applies only for RDs created by API.

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -107,6 +118,47 @@ const WorkspaceListTable = () => {
...
+    ) {
+      if (action === 'edit' && isValidEditType(workspace)) {
+        return true;
+      } else if (action === 'delete' && isValidDeleteType(workspace)) {
+        return true;
+      }
+    } else {
+      return false;
+    }
```
karelhala, line 159: Let's add default return statement here
```suggestion
    }
    return false;
```

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -198,6 +254,16 @@ const WorkspaceListTable = () => {
...
 
+  useEffect(() => {
+    chrome.getUserPermissions().then((permissions) => {
+      setUserPermissions(
+        permissions.find((obj) => {
+          return obj.permission === 'inventory:groups:write';
+        }),
+      );
+    });
+  }, []);
```
karelhala, line 265: Let's add `chrome` to dependency array otherwise on token refresh this wouldn't be calculated
```suggestion
  }, [chrome]);
```

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -198,6 +254,16 @@ const WorkspaceListTable = () => {
     dispatch(fetchWorkspaces());
   }, [dispatch]);
 
+  useEffect(() => {
+    chrome.getUserPermissions().then((permissions) => {
+      setUserPermissions(
+        permissions.find((obj) => {
+          return obj.permission === 'inventory:groups:write';
+        }),
```
karelhala, line 262: Let's add `inventory:groups:*` in here as well.
```suggestion
        permissions.find(({ permission }) => ['inventory:groups:write', 'inventory:groups:*'].includes(permission)),
```

karelhala (commented):
Looking really solid! Just a few code cleanup things.

src/smart-components/workspaces/WorkspaceListTable.tsx
karelhala: Let's try adding `chrome.getUserPermissions`. This funcion is re-initialized when new token is retrieved.

apinkert: @karelhala that worked! Just updated the PR.

src/smart-components/workspaces/WorkspaceListTable.tsx
```diff
@@ -143,14 +205,18 @@ const WorkspaceListTable = () => {
...
                 },
-                { title: <Divider component="li" key="divider" /> },
                 {
                   title: 'Delete workspace',
                   onClick: () => {
                     handleModalToggle([workspace]);
                   },
-                  isDisabled: workspace.children && workspace.children.length > 0,
+                  isDisabled: (workspace.children && workspace.children.length > 0) || !canModify(workspace, 'delete'),
                   isDanger: !(workspace.children && workspace.children.length > 0),
```
karelhala, line 220: We should also change this slightly so the disabled button is not of `danger` type.
```suggestion
                  isDanger: !(workspace.children && workspace.children.length > 0) && canModify(workspace, 'delete'),
```

PR #1852, opened 2025-06-06
fix(common-auth): current user cannot be disabled

karelhala:
### Description

If logged in user disables itself and is last org admin they have to contact redhat in order to re-enable the account back. API doesn't have a check for this for now and so users could block themselves from accessing any redhat app. This PR checks if username is same as currently logged in user and if so disable the option to disable / enable user activity.

[RHCLOUD-40473](https://issues.redhat.com/browse/RHCLOUD-40473)

---

#### Before:

Screenshot 2025-06-06 at 11 01 33

#### After:
Screenshot 2025-06-06 at 11 01 03

---

### Checklist ☑️

##
- [-] _(Optional) QE: Needs QE attention (OUIA changed, perceived impact to tests, no test coverage)_
- [-] _(Optional) QE: Has been mentioned_
- [-] _(Optional) UX: Needs UX attention (end user UX modified, missing designs)_
- [-] _(Optional) UX: Has been mentioned_
##


karelhala:
@Hyperkid123 good point on the tooltip. The option to mark user as org admin or not is disabled as well. Let's see if customers will want some tooltip or not.

FYI I've changed the logic as username felt too britle (users can change their username) so I'm comparing external source ID and account ID.

PR #1870, opened 2025-07-11
Increase viewport width

karelhala:
### Description

When we are running tests cypress defaults to smaller screen, that breaks our flow since some elements are hidden or changed. This PR fixes such issue by setting direct width to 1300px. There's also flaky test in users table where it tries to access ouiaId which is not present on the screen that's because we are using users table at one test and in other we are using users and groups. This PR changes from ouia ID to aria-label so we are testing directly what user can see and interact with.


riccardo-forina:
@karelhala is this still relevant? Is it worth a rebase?

PR #1897, opened 2025-08-12
Add initial workspace roles table

apinkert:
For [RHCLOUD-41321](https://issues.redhat.com/browse/RHCLOUD-41321). Since the WS roles API isn't ready, we're just displaying all groups via a `fetchGroups` call. This table replaces the current table with added drawer capabilities to shows the roles and users associated with a user group.

https://github.com/user-attachments/assets/9b3588cd-b1cd-49c1-83a9-ff5f0b1e325a

## Summary by Sourcery

Introduce an enhanced RoleAssignmentsTable for workspace details with sorting, filtering, and interactive drawers showing group-specific roles and users; integrate Redux-based data fetching and update WorkspaceDetail and Storybook accordingly.

New Features:
- Add clickable rows in RoleAssignmentsTable that open a drawer with Users and Roles tabs
- Enable sorting and filtering of user groups directly within the RoleAssignmentsTable
- Fetch group members and roles via Redux actions when a drawer is opened

Enhancements:
- Replace the previous table implementation with a PatternFly DataView-based table supporting bulk selection, pagination, and filters
- Extend WorkspaceDetail to manage sort, direction, and filter state and pass them to the fetchGroups API call
- Update Storybook stories to provide a mocked Redux store, include new sorting and filtering controls, and adjust play tests to validate skeleton loads, headers, and interactions

Tests:
- Enhance Storybook play functions to verify skeleton states, header presence, and new table behaviors

PR #1905, opened 2025-08-21
Add parent WS table for role assignments

apinkert:
For [RHCLOUD-41322](https://issues.redhat.com/browse/RHCLOUD-41322) and a follow up to https://github.com/RedHatInsights/insights-rbac-ui/pull/1897. This is using the `fetchGroups` data like the other table since the API is not ready at the moment

https://github.com/user-attachments/assets/110680f5-a069-4956-94b2-0eb1f49f4d16

## Summary by Sourcery

Add a new parent workspace role assignments table with inheritance support and integrate it into the workspace detail UI

New Features:
- Introduce ParentRoleAssignmentsTable component to display user groups and role assignments inherited from parent workspaces, complete with pagination, sorting, filtering, bulk selection, and drawer interactions

Enhancements:
- Extend GroupDetailsDrawer and RoleAssignmentsTable with an optional "Inherited from" column using GroupWithInheritance and RoleWithInheritance types
- Embed the new ParentRoleAssignmentsTable under the “Roles assigned in parent workspaces” tab in WorkspaceDetail and update related tab titles and filter parameters
- Extract EmptyTable into a reusable component for consistent empty state handling

Documentation:
- Add internationalization messages for new tab titles (“Roles assigned in this workspace”, “Roles assigned in parent workspaces”) and the "Inherited from" column
- Provide comprehensive Storybook stories demonstrating loading, empty state, filtering, pagination, sorting, and drawer interactions for the ParentRoleAssignmentsTable

### DIFF
--- a/src/features/workspaces/workspace-detail/WorkspaceDetail.tsx
+++ b/src/features/workspaces/workspace-detail/WorkspaceDetail.tsx
@@ -364,7 +364,6 @@ export const WorkspaceDetail = () => {
                       onSetFilters={parentOnSetFilters}
                       clearAllFilters={parentClearAllFilters}
                       ouiaId="parent-role-assignments-table"
-                      showInheritance={true}
                     />
                   )}
                 </div>
--- a/src/features/workspaces/workspace-detail/components/RoleAssignmentsTable.tsx
+++ b/src/features/workspaces/workspace-detail/components/RoleAssignmentsTable.tsx
@@ -17,6 +17,10 @@ import messages from '../../../../Messages';
 import { GroupDetailsDrawer, GroupWithInheritance } from './GroupDetailsDrawer';
 import { EmptyTable } from './EmptyTable';
 
+const isGroupWithInheritance = (group: Group | GroupWithInheritance): group is GroupWithInheritance => {
+  return 'inheritedFrom' in group && group.inheritedFrom !== undefined;
+};
+
 interface RoleAssignmentsTableProps {
   // Data props
   groups: Group[] | GroupWithInheritance[];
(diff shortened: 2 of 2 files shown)


src/features/workspaces/workspace-detail/WorkspaceDetail.tsx
```diff
@@ -236,7 +238,22 @@ export const WorkspaceDetail = () => {
...
+                      page={page}
+                      perPage={perPage}
+                      onSetPage={onSetPage}
+                      onPerPageSelect={onPerPageSelect}
+                      sortBy={sortBy}
+                      direction={direction}
+                      onSort={onSort}
+                      filters={filters}
+                      onSetFilters={onSetFilters}
+                      clearAllFilters={clearAllFilters}
```
karelhala, line 254: We should keep these props in the `ParentRoleAssignmentsTable` same for the `RoleAssignmentsTable`. These are state props of the table we don't want to share them between these 2 tables.

src/features/workspaces/workspace-detail/components/ParentRoleAssignmentsTable.tsx
karelhala, line 1: Can we reuse `RoleAssignmentsTable`? It has the same columns (apart from inherited from).

src/features/workspaces/workspace-detail/WorkspaceDetail.tsx
karelhala: If we're going to reuse `RoleAssignmentsTable` we should have two separate state values for each table.

karelhala (approved):

PR #2290, opened 2026-04-07
V2 defaultaccess issues

InsaneZein:
#### New Safeguards & UI Components

- Default Members Alert - Informs users that all users/org admins are automatically members of default groups

- Service Accounts Alert - Security compliance notice explaining service accounts are excluded from default groups

- Info Popover - Icons in the groups table explaining inherited roles for all default groups

- Reset Warning Modal - Confirmation dialog when attempting to restore a customized default group to system defaults

#### Edit Flow Enhancements

- Danger alert with acknowledgment checkbox prevents accidental customization of unmodified default groups

- Info alert with "Restore to default" button for already-customized groups

- Success notification after default group customization completes

---

### Attention needed
- [ ] _(Optional) QE: notable impact on test coverage or OUIA IDs changed_
- [ ] _(Optional) UX: end-user UX modified, designs may need sign-off_

## Summary by CodeRabbit

* **New Features**
  * Visual indicators and info popovers for platform/admin default groups
  * Alerts for default-group member and service-account behavior
  * Warning modal and restore flow for resetting customized default groups
* **Bug Fixes / UX**
  * Stronger safeguards when editing default groups (disabled fields, acknowledgement, and contextual actions)
* **Documentation / Stories**
  * Storybook coverage added for all new components and default-group scenarios

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -1981,6 +1981,16 @@ export default defineMessages({
     description: 'Users inherited roles message',
     defaultMessage: 'This group contains the roles that all users in your organization inherit by default.',
   },
+  defaultGroupInfoAriaLabel: {
+    id: 'defaultGroupInfoAriaLabel',
+    description: 'Aria label for default group info popover button',
+    defaultMessage: 'Default group information',
+  },
+  adminDefaultGroupInfoAriaLabel: {
+    id: 'adminDefaultGroupInfoAriaLabel',
+    description: 'Aria label for admin default group info popover button',
+    defaultMessage: 'Admin default group information',
+  },
   invalidGroup: {
     id: 'invalidGroup',
     description: 'Invalid group message',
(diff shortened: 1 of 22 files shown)


riccardo-forina (commented):
Good direction overall — UUID-based detection is the right fix, and the new components are well-scoped. Several issues need to be addressed before merge.

---

### 🔴 Critical

**1. Rename-only silently no-ops for unmodified default groups**

`handleSubmit` skips `updateGroupMutation` for `isUnmodifiedPlatformDefault` groups (the comment assumes the clone happens via member/SA changes). But if the user *only* changes the name and no members/SAs, no API call is made at all — the form just navigates back silently with no change applied and no error shown.

Either the name field should be disabled for default groups (see #4 below), or the rename path needs to be handled explicitly.

---

**2. Members/service accounts editing may be broken for unmodified default groups**

We verified in the backend code (`view.py:825`) that `protect_special_groups("add principals", group, additional="platform_default")` blocks both `system=True` and `platform_default=True` groups from having principals added. The backend team has indicated a fix has been merged to stage — **please confirm this is the expected new behavior before merge**, as the entire edit flow for default groups depends on it.

---

### 🟠 Major

**3. `danger` alert variant should be `warning`**

`EditUserGroup.tsx` uses `variant="danger"` for the pre-edit alert on unmodified default groups. In V1, the equivalent gate was a `WarningModal` (warning severity), and the post-change alert used `variant="info"`. `danger` is semantically reserved for errors or actively destructive operations in progress — a proactive "heads up" before the user has done anything warrants `variant="warning"`.

**4. Name/description field should be disabled for default groups**

In V1, the edit action (rename/redescribe) was completely hidden for `platform_default` and `admin_default` groups — the actions dropdown was `null` for those groups (`GroupHeader.tsx:81`). The V2 edit form presents the name field as editable for default groups, which has no V1 precedent, causes the silent no-op bug above, and the backend also explicitly blocks renaming customized default groups (`restrict_custom_default_group_renaming` in `view.py:263`).

Recommended fix: disable the name (and description) fields for `isUnmodifiedPlatformDefault` groups using DDF's `isDisabled` + `helperText` to explain why and that renaming will be possible after the group is customized.

**5. Double success toast on restore**

`handleRestoreConfirm` calls `deleteGroupMutation.mutateAsync` then manually fires a success notification. `useDeleteGroupMutation` already calls `notify('success', ...)` in its `onSuccess` handler — users will see two toasts. Suppress the mutation's default notification or skip the manual one.

**6. UUID queries not reflected in `isLoading`**

`useWorkspaceGroups`, `useWorkspaceInheritedGroups`, and `useOrgGroups` each fetch `systemGroup` and `adminGroup` UUIDs but only expose `isLoading` from the bindings query. If bindings resolve before the UUID queries, default groups are emitted as ordinary rows — skipping alerts, popovers, and safeguards until the UUID queries settle. Include `systemGroup.isLoading || adminGroup.isLoading` in each hook's returned `isLoading`.

---

### 🟡 Minor

**7. Full page re-render on checkbox toggle**

`defaultGroupAcknowledged` state lives in the parent `EditUserGroup` component. Toggling the checkbox re-renders the entire page including `FormRenderer` and all data-fetching hooks. Move the checkbox state into `CustomFormTemplate` itself so only the template re-renders.

It's also worth exploring whether the checkbox can live inline in the footer action group area (alongside Submit/Cancel) rather than as a separate block above it — this would tighten the visual relationship between the acknowledgement and the action.

**8. `CustomFormTemplate` defined via `useCallback` with a hook inside**

`CustomFormTemplate` is created with `useCallback` but calls `useFormApi()` internally. When passed as a prop to `FormRenderer`, React may not consistently treat it as a stable component boundary, making the hook call potentially unstable across renders. Extract it as a named component (`function CustomFormTemplate(props) { ... }`) outside the parent component.

**9. Hardcoded English strings (not i18n)**

Several notification strings are raw English instead of using `intl.formatMessage`:
- `"Cannot edit Default admin access group"` / `"The Default admin access group cannot be edited..."` (`useEffect` redirect)
- `"Default access group customized"` / `"The group has been customized..."` (post-save notification)
- `"Error restoring group"` / `"There was an error restoring..."` (restore error handler)

**10. `admin_default` redirect has a race condition**

The `useEffect`-based redirect fires after render, causing a brief flash of the edit form before navigating away. Consider moving this guard earlier (e.g. render `null` or a spinner while `group` is loading, then redirect synchronously on the first render where `group.admin_default` is known).

**11. Duplicate popover `id` attributes** *(also flagged by CodeRabbit)*

`useUserGroupsTableConfig.tsx` generates `id={`default${group.admin_default ? '-admin' : ''}-group-popover`}` — only two possible values regardless of how many rows exist. Append `group.uuid`:
```tsx
id={`default${group.admin_default ? '-admin' : ''}-group-popover-${group.uuid}`}
```

**12. `DefaultInfoPopover` is not keyboard accessible** *(also flagged by CodeRabbit)*

The trigger is a bare `OutlinedQuestionCircleIcon` — not focusable, no accessible name, and `triggerAction="hover"` only means keyboard/touch users can never open it. Wrap in a `<Button variant="plain" aria-label="...">`.

**13. `PageSection` has a manually added margin class**

```tsx
<PageSection ... className="pf-v6-u-m-lg-on-lg" isWidthLimited>
```
The `pf-v6-u-m-lg-on-lg` class appears to have been added manually — remove it if not intentional.

**14. Missing second arg to `createGroupMembersHandlers` in stories** *(flagged by CodeRabbit)*

The new `DefaultPlatformGroup` and `DefaultAdminGroup` stories in `GroupDetailsUsersView.stories.tsx` call `createGroupMembersHandlers({ ... })` with one argument; existing stories pass two (user map + service account map). This may cause runtime errors in the story.

InsaneZein:
PR Feedback addressed:
  - Field disabling with helper text for unmodified default groups
  -  Alert severity changed from "danger" to "warning"
  -  State optimization (moved checkbox to CustomFormTemplate)
  -  Duplicate notifications fixed with deferSuccessSideEffects
  -  UUID query loading states included in all hooks
  -  CustomFormTemplate extracted as named component
  -  All hardcoded strings replaced with intl.formatMessage()
  -  Admin default redirect race condition fixed
  -  Unique popover IDs with UUID appended
  -  Keyboard accessibility added to DefaultInfoPopover
  -  Missing story arguments fixed

riccardo-forina:
We can't really edit the default access group anymore, we can only update the role bindings. A new ticket reflecting this will be created

PR #2310, opened 2026-04-16
fix(table-view): filter non-selectable rows in bulk select

aferd:
## Summary

- **Bug**: `TableView.handleBulkSelect` passed all rows to `onSelectAll` without filtering by `isRowSelectable`, causing "Select all" to include non-selectable rows (e.g., system/default roles)
- **Fix**: Filter rows through `isRowSelectable` before passing to `onSelectAll` callback
- **Test**: Added `BulkSelectRespectsIsRowSelectable` Storybook interaction test with static data verifying only selectable rows are included in bulk selection

RHCLOUD-46661

## Test plan

- [x] All 1019 Storybook tests passing (174 suites)
- [x] All 210 unit tests passing (14 suites)
- [x] ESLint + Prettier + circular dependency checks passing
- [ ] Verify in staging: navigate to V2 Roles table, click "Select all" — system/default roles should not be selected

## Summary by CodeRabbit

* **Bug Fixes**
  * Bulk select now respects per-row selectability so only selectable rows are included in bulk operations and counts reflect actual selections.

* **Tests**
  * Added a test that verifies bulk-select only affects selectable rows, including checkbox states and selected count.

* **Examples**
  * Added an interactive story demonstrating and validating bulk-select behavior with non-selectable rows.

### DIFF
--- a/src/shared/components/table-view/TableView.stories.tsx
+++ b/src/shared/components/table-view/TableView.stories.tsx
@@ -1150,7 +1150,13 @@ const BulkSelectIsRowSelectableTable: React.FC = () => {
       ariaLabel="Bulk select test table"
       ouiaId="BulkSelectTest"
       bulkActions={
-        tableState.selectedRows.length > 0 ? <span data-testid="selected-count">Selected: {tableState.selectedRows.length}</span> : undefined
+        tableState.selectedRows.length > 0 ? (
+          <>
+            <span data-testid="selected-count">Selected: {tableState.selectedRows.length}</span>
+            <span data-testid="selected-ids">{tableState.selectedRows.map((row) => row.id).join(',')}</span>
+            <span data-testid="selected-names">{tableState.selectedRows.map((row) => row.name).join(',')}</span>
+          </>
+        ) : undefined
       }
       {...tableState}
     />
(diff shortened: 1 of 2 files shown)


apinkert (approved):

riccardo-forina:
Did we check this against the case reported in the ticket, as extra validation?

aferd:
The fix was verified against the exact case from the ticket:

1. **Scenario**: Click "Select all" checkbox in the roles table when default/system roles (Compliance administrator, Inventory administrator, etc.) are present
2. **Before fix**: The "Select all" toggled ALL rows including default roles that have no individual checkbox, causing the bulk delete action to fail with an error
3. **After fix**: The `onSelectAll` handler in `TableView.tsx` now filters out rows where `isSelectable` is false — only custom (non-default) roles get selected. Default roles remain unaffected by bulk selection

The Storybook story `SelectAllWithNonSelectableRows` demonstrates the exact ticket scenario: a table with a mix of selectable and non-selectable rows. The play test verifies that "Select all" only checks the selectable rows.

riccardo-forina:
Does it check it visually only? Or does it check that the actual list of selected rows is the the right length and with the right content?

aferd:
Good question! The existing test verified the visual state (checkboxes checked/unchecked) and the count (`Selected: 3`), but didn't verify _which_ rows were in the `selectedRows` array.

I've enhanced the test to verify the actual content:

**Component change** — `BulkSelectIsRowSelectableTable` now renders `data-testid="selected-ids"` and `data-testid="selected-names"` elements showing the actual `selectedRows` data:
```tsx
<span data-testid="selected-ids">{tableState.selectedRows.map((row) => row.id).join(',')}</span>
<span data-testid="selected-names">{tableState.selectedRows.map((row) => row.name).join(',')}</span>
```

**Test assertions** — after bulk select, the test now verifies:
1. ✅ Selected IDs are exactly `1,3,5` (the non-system rows)
2. ✅ System row IDs `2` and `4` are NOT in the selection
3. ✅ Selected names are `Custom Role A,Custom Role C,Custom Role E`
4. ✅ System role names (`System Role B`, `System Role D`) are NOT in the selection
5. ✅ All 210 tests pass

So it checks both the visual state and the actual `selectedRows` content.

_Commit `e431a986` is ready locally — will push shortly._

PR #2311, opened 2026-04-16
fix(users): handle 403 in user detail drawer

aferd:
## Summary
- **Fixes RHCLOUD-46489**: UserViewer persona clicking another user's drawer triggered a full-page "Unauthorized" error because the groups API 403 propagated through QueryCache → ApiErrorContext → ApiErrorBoundary
- Added `meta.skipGlobalErrorHandler` support to QueryClientSetup so individual queries can opt out of the global error boundary
- User detail drawer's Groups and Roles tabs now handle 403 locally, showing a friendly "Permission needed" message with a lock icon instead of crashing the page
- Added `PermissionDenied` Storybook stories for both `UserDetailsGroupsView` and `UserDetailsRolesView`

## Test plan
- [ ] All 210 vitest tests pass
- [ ] Lint + prettier clean
- [ ] New `PermissionDenied` stories render correctly in Storybook
- [ ] Existing `APIError` and `NetworkFailure` stories still show "Unable to load" (non-403 errors unaffected)
- [ ] Manual verification: log in as UserViewer, click another user → drawer shows "Permission needed" in Groups tab instead of full-page error

## Summary by CodeRabbit

* **New Features**
  * Permission-gated User Details: Groups and Roles tabs show only when allowed; header-only view while permissions load; active tab auto-switches to Roles if Groups is unavailable; focused user prop is now required for the drawer content.

* **Tests**
  * Added stories for permission states: no access, groups-only, roles-only; tab assertions updated to role-based queries.

* **Documentation**
  * Clarified inline comments about permission handling in user-details components.

### DIFF
--- a/src/shared/components/QueryClientSetup.tsx
+++ b/src/shared/components/QueryClientSetup.tsx
@@ -64,17 +64,10 @@ const PRODUCTION_QUERY_OPTIONS = {
 
 /**
  * Creates a QueryClient with production settings and error handling.
- * Queries with `meta.skipGlobalErrorHandler = true` bypass the global error handler,
- * allowing components to handle errors locally (e.g., 403 in a drawer panel).
  */
 function createQueryClient(onError: (error: unknown) => void): QueryClient {
   return new QueryClient({
-    queryCache: new QueryCache({
-      onError: (error, query) => {
-        if (query.meta?.skipGlobalErrorHandler) return;
-        onError(error);
-      },
-    }),
+    queryCache: new QueryCache({ onError }),
     mutationCache: new MutationCache({ onError }),
     defaultOptions: PRODUCTION_QUERY_OPTIONS,
   });
@@ -85,12 +78,7 @@ function createQueryClient(onError: (error: unknown) => void): QueryClient {
  */
 function createTestQueryClient(onError: (error: unknown) => void): QueryClient {
   return new QueryClient({
-    queryCache: new QueryCache({
-      onError: (error, query) => {
-        if (query.meta?.skipGlobalErrorHandler) return;
-        onError(error);
-      },
-    }),
+    queryCache: new QueryCache({ onError }),
     mutationCache: new MutationCache({ onError }),
     defaultOptions: TEST_QUERY_OPTIONS,
   });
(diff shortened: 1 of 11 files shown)


riccardo-forina:
This wasn't the plan. What we had to do was checking for the permissions granted to the current user, and in case they had bo access to groups/roles listing, we should have hidden the tabs and disabled the drawer. This way the 403 error would not trigger at all

aferd:
Good call — reworked in 07a47c53.

**What changed:**
- `UserDetailsDrawer` now calls `useGroupsAccess()` and `useRolesAccess()` upfront to check `rbac_groups_read` / `rbac_roles_read` before rendering tabs
- If the user lacks a permission, the corresponding tab is hidden entirely — the query never fires, so 403 is never triggered
- If both permissions are denied, the drawer shows a "Permission needed" empty state
- Reverted the `skipGlobalErrorHandler` infrastructure from `QueryClientSetup`, `types.ts`, `groups.ts`, and `roleBindings.ts` — no longer needed
- Removed `PermissionDenied` stories from the individual view components; added `PermissionDenied`, `GroupsOnlyPermission`, and `RolesOnlyPermission` stories to the drawer instead
- 210 tests pass, lint clean

riccardo-forina:
The drawer should not show a permission denied screen. As for the parent page, if users don't have access to the groups/roles they should not see the tab at all

aferd:
Good point — pushed cebbacf0.

**What changed:**
- Removed the "Permission needed" EmptyState entirely — the drawer no longer shows any error screen when permissions are absent
- When the user lacks both `rbac_groups_read` and `rbac_roles_read`, the drawer shows only the user header (name, email, close button) with no tabs
- Consistent with how the parent page handles permissions: if you don't have access, the tab simply isn't there
- All 1021 Storybook tests pass, lint clean

The `NoPermissions` story demonstrates the new behavior.

riccardo-forina:
The suggestion from Coderabbit is worth doing

aferd:
Done in 2aaaa1ab — `UserDetailsDrawerContent` is now only rendered when `focusedUser` is defined. When the drawer is closed, an empty `DrawerPanelContent` is used as a placeholder, so `useGroupsAccess`/`useRolesAccess` hooks no longer fire on every page load.

`focusedUser` is now required (non-optional) in `UserDetailsDrawerContent`, which also simplified the internal guards. All 210 tests pass.

riccardo-forina (approved):

PR #2312, opened 2026-04-17
fix(workspaces): disable delete for workspaces with children

aferd:
## Summary

- Fixed the workspace detail page allowing delete on workspaces that have children — delete should only be available on bottom-level (leaf) workspaces
- `useWorkspaceDetailData` now computes `hasChildren` by checking if any workspace has `parent_id` matching the current workspace
- Threaded `hasChildren` through `WorkspaceDetailLayout` → `WorkspaceHeader` → `useWorkspaceActionItems` (matching the existing pattern in `WorkspaceListTable`)
- Added two Storybook stories verifying delete is disabled with children and enabled for leaf workspaces

RHCLOUD-46470

## Test plan

- [ ] Open a workspace detail page for a workspace that has child workspaces → Actions menu → "Delete workspace" should be **disabled**
- [ ] Open a workspace detail page for a leaf workspace (no children) → Actions menu → "Delete workspace" should be **enabled** (if user has delete permission)
- [ ] Verify the workspace list table still behaves correctly (unchanged)
- [ ] Run Storybook stories: `DeleteDisabledWithChildren` and `DeleteEnabledLeafWorkspace`

## Summary by CodeRabbit

* **Tests**
  * Added stories and test interactions to validate “Delete workspace” behavior for workspaces with and without children; made assertions resilient to async state and network timing.

* **Refactoring**
  * Switched deletion logic to rely on workspace identifiers and internal lookup of child workspaces.
  * Introduced a fail-closed behavior that disables delete while child presence is unknown.

* **Chores**
  * Wired request mocking for Storybook to stabilize story data.

### DIFF
--- a/src/v2/features/workspaces/components/WorkspaceHeader.tsx
+++ b/src/v2/features/workspaces/components/WorkspaceHeader.tsx
@@ -37,7 +37,7 @@ export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({ workspace, isL
   const showChildContextAlert = fromChildId && fromChildName;
 
   const actionItems = useWorkspaceActionItems({
-    workspaceId: workspace?.id ?? '',
+    workspaceId: workspace?.id,
     permissions,
     callbacks: actionCallbacks,
   });
--- a/src/v2/features/workspaces/components/WorkspaceHeader.tsx
+++ b/src/v2/features/workspaces/components/WorkspaceHeader.tsx
@@ -21,7 +21,7 @@ export interface WorkspaceHierarchyItem {
 }
 
 export interface WorkspaceHeaderProps {
-  workspace: (WorkspacesWorkspace & { children?: readonly unknown[] }) | null;
+  workspace: WorkspacesWorkspace | null;
   isLoading: boolean;
   workspaceHierarchy: WorkspaceHierarchyItem[];
   permissions?: WorkspacePermissions;
@@ -37,7 +37,7 @@ export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({ workspace, isL
   const showChildContextAlert = fromChildId && fromChildName;
 
   const actionItems = useWorkspaceActionItems({
-    workspace: workspace ?? ({ id: '', name: '' } as WorkspacesWorkspace),
+    workspaceId: workspace?.id ?? '',
     permissions,
     callbacks: actionCallbacks,
   });
--- a/src/v2/features/workspaces/components/WorkspaceHeader.tsx
+++ b/src/v2/features/workspaces/components/WorkspaceHeader.tsx
@@ -26,9 +26,17 @@ export interface WorkspaceHeaderProps {
   workspaceHierarchy: WorkspaceHierarchyItem[];
   permissions?: WorkspacePermissions;
   actionCallbacks: WorkspaceActionCallbacks;
+  hasChildren?: boolean;
 }
 
-export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({ workspace, isLoading, workspaceHierarchy, permissions, actionCallbacks }) => {
+export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({
+  workspace,
+  isLoading,
+  workspaceHierarchy,
+  permissions,
+  actionCallbacks,
+  hasChildren = false,
+}) => {
   const intl = useIntl();
   const [searchParams] = useSearchParams();
 
@@ -40,6 +48,7 @@ export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({ workspace, isL
     workspace: workspace ?? ({ id: '', name: '' } as WorkspacesWorkspace),
     permissions,
     callbacks: actionCallbacks,
+    hasChildren,
   });
 
   const pageBreadcrumbs = useMemo(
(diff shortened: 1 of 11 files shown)


riccardo-forina:
This actually would be better fixed by making the check on the children mandatory. It never makes sense not to check the children when evaluating the delete check

aferd:
Good point — reworked in 894f3708.

**What changed:**
- `useWorkspaceActionItems` now computes `hasChildren` internally from `workspace.children` instead of accepting it as an optional prop
- Removed the `hasChildren` prop threading through the entire detail page hierarchy (tabs → layout → header → hook)
- `useWorkspaceDetailData` enriches the workspace object with its `children` array, so the hook always has the data it needs
- The original bug root cause (optional `hasChildren` defaulting to `false`) is now impossible — the check always happens

210 tests pass, lint clean.

src/v2/features/workspaces/workspace-detail/useWorkspaceDetailData.ts
```diff
@@ -36,7 +36,12 @@ export function useWorkspaceDetailData(): WorkspaceDetailData {
   const { workspaceId = '' } = useParams<{ workspaceId: string }>();
   const { workspaces, status } = useWorkspacesWithPermissions();
 
-  const workspace = useMemo<WorkspaceWithPermissions | null>(() => workspaces.find((ws) => ws.id === workspaceId) ?? null, [workspaces, workspaceId]);
+  const workspace = useMemo(() => {
+    const found = workspaces.find((ws) => ws.id === workspaceId) ?? null;
+    if (!found) return null;
+    const children = workspaces.filter((ws) => ws.parent_id === workspaceId);
+    return { ...found, children };
+  }, [workspaces, workspaceId]);
```
riccardo-forina: This block should live inside the hook. The hook should not ask for the computed Workspace object but only for a workspace id. We are leaking complexity and allowing bad actors to bypass the children check by passing an empty array

riccardo-forina (changes requested):

src/v2/features/workspaces/workspace-detail/useWorkspaceDetailData.ts
aferd: Good call — pushed e82a382a.

**What changed:**
- `useWorkspaceActionItems` now takes `workspaceId: string` instead of the full workspace object
- The hook calls `useWorkspacesQuery()` internally and checks `allWorkspaces.some(ws => ws.parent_id === workspaceId)` — callers can't bypass the children check
- Fail-closed: delete is disabled when workspace data is loading, errored, or workspace ID is empty
- Removed `WorkspacesWorkspace` from the hook's type imports entirely — it only needs a string
- Cleaned up `useWorkspaceDetailData` — no longer enriches workspace with children
- All callers (`WorkspaceHeader`, `WorkspaceListTable`, stories) pass `workspace.id` instead of the full object

All 210 tests pass, lint clean.

src/v2/features/workspaces/components/useWorkspaceActionItems.ts
```diff
@@ -23,22 +23,22 @@ export interface WorkspaceActionItem {
...
+export function useWorkspaceActionItems({ workspaceId, permissions, callbacks }: UseWorkspaceActionItemsParams): WorkspaceActionItem[] {
   const intl = useIntl();
   const hasM4Flag = useWorkspacesFlag('m4');
   const perms = permissions ?? EMPTY_PERMISSIONS;
 
+  // Children check is computed internally — callers cannot bypass it.
+  // Fail-closed: disable delete when workspace data is unavailable.
+  const { data: workspaceList, isLoading, isError } = useWorkspacesQuery();
+  const allWorkspaces = workspaceList?.data ?? [];
+  const hasChildren = isLoading || isError || !workspaceId || allWorkspaces.some((ws) => ws.parent_id === workspaceId);
```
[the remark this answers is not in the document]
riccardo-forina: Not a problem, the actual limit is 120

src/v2/features/workspaces/components/WorkspaceHeader.tsx
```diff
@@ -37,7 +37,7 @@ export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({ workspace, isL
   const showChildContextAlert = fromChildId && fromChildName;
 
   const actionItems = useWorkspaceActionItems({
-    workspace: workspace ?? ({ id: '', name: '' } as WorkspacesWorkspace),
+    workspaceId: workspace?.id ?? '',
```
riccardo-forina: This should allow for undefined, so to use the undefined id to disable the query in the hook. Else we risk calling an API with an empty id

riccardo-forina (changes requested):

src/v2/features/workspaces/components/WorkspaceHeader.tsx
aferd: Good call — pushed c04266ef.

**What changed:**
- `useWorkspaceActionItems` now accepts `workspaceId: string | undefined`
- The `useWorkspacesQuery()` call uses `enabled: !!workspaceId` — query is disabled when the ID is undefined, preventing API calls with empty/missing IDs
- Removed the `?? ''` fallback from all callers (`WorkspaceHeader`, `WorkspaceListTable`, `WorkspaceActions.stories`)
- Fail-closed behavior preserved: `!workspaceId` still disables delete in the `hasChildren` computation

210 tests pass, lint clean.

riccardo-forina (approved):

